PR #1, opened 2015-10-15
Initial structure

stdweird:
Largely based on vsc-base 2.4.4
* insert `base_dir` in `sys.path`
* tries to reload the `DEFAULT_TEST_SUITE` namespace
* `__main__` is the `setup.py` for this repo


stdweird:
@JensTimmerman who wants to setup jenkins?

test/simple.py
```diff
@@ -0,0 +1,9 @@
+from unittest import TestCase, TestLoader
+import vsc.install.shared_setup
+
+class ImportTest(TestCase):
+    """Dummy class to prove the shared_setup import works"""
+
+    def test_importok(self):
+        """It's ok if we get here"""
+        self.assertTrue(True, msg='import vsc.install.shared_setup was success')
```
boegel: do the import here in the test itself then?

test/__init__.py
```diff
@@ -0,0 +1 @@
+
```
boegel: keep this *empty*?

setup.py
```diff
@@ -0,0 +1 @@
+lib/vsc/install/shared_setup.py
```
boegel: how does this work? add a comment to clarify?

lib/vsc/install/shared_setup.py
```diff
@@ -0,0 +1,548 @@
...
+    PACKAGE = {
+        'name': 'vsc-install',
+        'version': '1.0.0',
+        'author': [sdw, ag, jt],
+        'maintainer': [sdw, ag, jt],
+        'packages': ['vsc', 'vsc.install'],
+        'zip_safe': True,
+    }
+
+    action_target(PACKAGE, urltemplate=URL_GH_HPCUGENT)
```
boegel: might as well make `URL_GH_HPCUGENT` the default value for `urltemplate` (see https://github.com/hpcugent/vsc-base/pull/203)

boegel:
@stdweird: can you edit the description with an overview of what you changed to files in `vsc/install` that already existed in `vsc-base`, for future reference?

boegel:
@stdweird: wrong `test` namespace is being picked up...

setup.py
stdweird: what do you mean? it's a symlink to shared_setup.py

boegel: ah, I missed that, sorry, nvm :)

stdweird:
The vsc.fancylogger hack is required to provide vsc.fancylogger without vsc.utils
As as such usage will log to syslog, we can track it and hunt it down before we remove it again.

lib/vsc/fancylogger.py
```diff
@@ -0,0 +1,747 @@
...
+# vsc-base is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Library General Public License for more details.
+#
+# You should have received a copy of the GNU Library General Public License
+# along with vsc-base. If not, see <http://www.gnu.org/licenses/>.
+# #
+"""
+This module implements a fancy logger on top of python logging
```
JensTimmerman: add a line here to indicate that this file is deprecated and should not be used, use `vsc.utils.fancylogger` instead

boegel:
Wait, what? We're shipping fancylogger here too now? Haha, no, please no.

stdweird:
unless you can propose another clean migration path to finally get rid of `vsc.fancylogger`...

boegel:
Copying it is just nuts, you're making things worse with having two copies of `fancylogger.py` around and assuming that people will never confuse them.

Why not simply use this in `vsc/fancylogger.py`:

```python
from vsc.utils.fancylogger import *
```

If you import that, and the import in it fails, then you don't have fancylogger installed, so that's OK?

That, together with your syslog thingie so we can track down the usage, that's all that should be in `vsc/fancylogger.py`, not actual code.

Also, why exactly does this need to be in this repo? fancylogger is in vsc-base, so any migration path issues should be dealt with there, not here?

stdweird:
vsc-install ships `vsc` namespace (in rpms!), so it has to be here. but the import * trick is indeed a lot better

lib/vsc/fancylogger.py
```diff
@@ -0,0 +1,17 @@
+"""
+vsc.fancylogger must be removed:
```
boegel: mention that `vsc.utils.fancylogger` should be used instead

test/__init__.py
```diff
@@ -0,0 +1 @@
+# Intentionally left empty
```
boegel: why do this (i.e., include a comment) here, and not in `vsc/install/__init__.py` where people may actually wonder why the `extend_path` thing isn't there (it shouldn't be, and it should be very clear that's not a mistake)

stdweird: because i added it to the wrong file :smile:

stdweird:
There is still an issue with the rpm generation and the python prefixing.

lib/vsc/install/shared_setup.py
```diff
@@ -0,0 +1,588 @@
...
+
+# Put unittests under this directory
+DEFAULT_TEST_SUITE = 'test'
+
+URL_GH_HPCUGENT = 'https://github.com/hpcugent/%(name)s'
+URL_GHUGENT_HPCUGENT = 'https://github.ugent.be/hpcugent/%(name)s'
+
+RELOAD_VSC_MODS = False
+
+VERSION = '1.0.0'
```
boegel: make this `0.9.0`, bump to `1.0` when you kick out `vsc/utils`

stdweird:
@boegel done

PR #3, opened 2015-10-20
Fix rpm packaging issue

stdweird:


lib/vsc/install/shared_setup.py
```diff
@@ -648,7 +648,7 @@ def action_target(target, setupfn=setup, extra_sdist=[], urltemplate=None):
         'zip_safe': True,
         'install_requires': ['setuptools'],
         'setup_requires': ['setuptools'],
-        'excluded_pkg_rpm': [], # vsc-install ships vsc package (the vsc package is removed by default)
+        'excluded_pkgs_rpm': [], # vsc-install ships vsc package (the vsc package is removed by default)
```
boegel: magic string problem? ;)

stdweird: yep

PR #4, opened 2015-10-22
Add vsc_sdist code to ship and integrate shared_setup with sdist

stdweird:


stdweird:
@JensTimmerman this produces a tarball with sdist, when unpacked, can be used to run `easy_intall .`
(cd-ing in the untarred dir, and running `python setup.py easy_install .` fails).

stdweird:
hmm, cd'ing in the dir and python setup.py easy_install works on el6...
would need to test this on other machines and with a push to pypi
@boegel any objections to merge and push to pypi?

boegel:
@stdweird: I guess I need some context here... What are you trying to solve with this?

And what happened to your plans to merge `vsc-install` back into `vsc-base`?

I'm not strongly opposing pushing this to PyPI, but bumping the `vsc-base` release to something that requires `vsc-install`, and can't be installed without having `vsc.install` already available somehow is a bigger issue, since that's going to break the EasyBuild bootstrap...

stdweird:
this PR allows to fix that, for all repos with shared_setup. it basically injects shared_setup in the sdist phase, so we don't need it cluttered around in the code repos, but people using e.g. pypi or sdist tarballs should be able to run the bootstrap (given they have setuptools installed). with this PR, there's no reason to merge (well, unless this is not working :smile: )

lib/vsc/install/shared_setup.py
```diff
@@ -470,18 +561,19 @@ def generate_packages(extra=None, exclude=None):
...
+        "install_scripts": vsc_install_scripts,
+        "sdist": vsc_sdist,
         "test": VscTestCommand,
     },
-    'test_suite': DEFAULT_TEST_SUITE,
-    'setup_requires' : ['setuptools', 'vsc-install >= %s' % VERSION],
+    'download_url': '',
+    'package_dir': {'': DEFAULT_LIB_DIR},
     'packages': generate_packages(),
+    'setup_requires' : ['setuptools', 'vsc-install >= %s' % VERSION],
```
boegel: does this imply that vsc-install requires vsc-install?

boegel: ah, no, this is for anything else, e.g. vsc-base and 'up', right?

lib/vsc/install/shared_setup.py
```diff
@@ -524,9 +616,18 @@ def sanitize(name):
...
+    Add some fields
+
+    Remove sdist vsc class with '"vsc_sdist": False' in target
+    """
     new_target = {}
     new_target.update(SHARED_TARGET)
+
+    vsc_sdist = target.pop('vsc_sdist', True)
+    if not vsc_sdist:
+        new_target['cmdclass'].pop('sdist')
```
boegel: do you need to check whether `sdist` is something else than the `vsc_sdist` class?

lib/vsc/install/shared_setup.py
```diff
@@ -156,6 +165,88 @@ def remove_extra_bdist_rpm_files(pkgs=None):
 
     return res
 
+class vsc_sdist(sdist):
```
boegel: class name should be CamelCase, so `VscSdist`? or do you want to stick to the setuptools style of class names?

lib/vsc/install/shared_setup.py
```diff
@@ -156,6 +165,88 @@ def remove_extra_bdist_rpm_files(pkgs=None):
...
+    """
+    Upon sdist, add this vsc.install.shared_setup to the sdist
+    and modifed the shipped setup.py to be able to use this
+    """
+    def _mod_setup_py(self, base_dir, external_dir, new_shared_setup):
+        """Modify the setup.py in the distribution directory"""
+
+        # re-copy setup.py, to avoid hardlinks (code from setupstools)
```
boegel: where did you steal this from exactly (file/function in setuptools?)

lib/vsc/install/shared_setup.py
```diff
@@ -156,6 +165,88 @@ def remove_extra_bdist_rpm_files(pkgs=None):
...
+
+        # insert sys.path hack
+        before = reg.start()
+        # no indentation
+        code = code[:before] + NEW_SHARED_SETUP_HEADER_TEMPLATE % (new_shared_setup, VERSION, external_dir) + code[before:]
+
+        # replace 'vsc.install.shared_setup' -> new_shared_setup
+        code = re.sub(r'vsc\.install\.shared_setup', new_shared_setup, code)
+        # replace 'from vsc.install import shared_setup' -> import new_shared_setup as shared_setup
+        code = re.sub(r'from\s+vsc.install\s+import\s+shared_setup', 'import %s as shared_setup', code)
```
boegel: what if `shared_setup` was imported like this in the original `setup.py`?

```
from vsc.install import shared_setup as foo
```

this regex will break that?

lib/vsc/install/shared_setup.py
```diff
@@ -156,6 +165,88 @@ def remove_extra_bdist_rpm_files(pkgs=None):
...
+
+        ext_dir = os.path.join(base_dir, external_dir)
+        os.mkdir(ext_dir)
+
+        dest = os.path.join(ext_dir, '%s.py' % new_shared_setup)
+        log.error('inserting shared_setup as %s' % dest)
+        try:
+            source_code = inspect.getsource(sys.modules[__name__])
+        except:
+            raise Exception("sdist requires access shared_setup source")
```
boegel: `access to`?
`shared_setup.py`?

lib/vsc/install/shared_setup.py
```diff
@@ -156,6 +165,88 @@ def remove_extra_bdist_rpm_files(pkgs=None):
...
+    def _add_shared_setup(self, base_dir, external_dir, new_shared_setup):
+        """Create the new shared_setup in distribution directory"""
+
+        ext_dir = os.path.join(base_dir, external_dir)
+        os.mkdir(ext_dir)
+
+        dest = os.path.join(ext_dir, '%s.py' % new_shared_setup)
+        log.error('inserting shared_setup as %s' % dest)
+        try:
+            source_code = inspect.getsource(sys.modules[__name__])
```
boegel: hmm, maybe use `'shared_setup'` rather than `__name__`?
not sure if it makes a big difference...

lib/vsc/install/shared_setup.py
```diff
@@ -156,6 +165,88 @@ def remove_extra_bdist_rpm_files(pkgs=None):
...
+        except:
+            raise Exception("sdist requires access shared_setup source")
+        try:
+            fh = open(dest, 'w')
+            fh.write(source_code)
+            fh.close()
+        except:
```
boegel: `... as err`, and mention what went wrong in the `raise` below? don't hide stuff...

lib/vsc/install/shared_setup.py
```diff
@@ -156,6 +165,88 @@ def remove_extra_bdist_rpm_files(pkgs=None):
...
+        """Create the new shared_setup in distribution directory"""
+
+        ext_dir = os.path.join(base_dir, external_dir)
+        os.mkdir(ext_dir)
+
+        dest = os.path.join(ext_dir, '%s.py' % new_shared_setup)
+        log.error('inserting shared_setup as %s' % dest)
+        try:
+            source_code = inspect.getsource(sys.modules[__name__])
+        except:
```
boegel: same here, catch `err`, use it below

lib/vsc/install/shared_setup.py
```diff
@@ -156,6 +165,88 @@ def remove_extra_bdist_rpm_files(pkgs=None):
...
+            raise Exception("sdist requires access shared_setup source")
+        try:
+            fh = open(dest, 'w')
+            fh.write(source_code)
+            fh.close()
+        except:
+            raise Exception("Failed to write new_shared_setup source to %s" % dest)
+
+    def make_release_tree(self, base_dir, files):
+        log.error("sdist make_release_tree original base_dir %s files %s" % (base_dir, files))
```
boegel: docstring pls

lib/vsc/install/shared_setup.py
```diff
@@ -156,6 +165,88 @@ def remove_extra_bdist_rpm_files(pkgs=None):
...
+        if os.path.exists(base_dir):
+            # no autocleanup?
+            # can be a leftover of earlier crash/raised exception
+            raise Exception("base_dir %s present. Please remove it" % base_dir)
+
+        sdist.make_release_tree(self, base_dir, files)
+
+        if __name__ == '__main__':
+            log.error('running shared_setup as main, not adding it to sdist')
+            return
```
boegel: don't return inline, use `else` instead, and only return below (which is not needed, `return` is there implicitely?

boegel:
reviewed, may work...

what kind of error(s) will we run into if someone tries to install `vsc-base` from git directly, or via `master.tar.gz` straight from GitHub? can we make sure a clear error message is produced then (something like `"only install from PyPI, or make sure a recent `vsc.install` is available"`)?

lib/vsc/install/shared_setup.py
stdweird: yes, see vsc-install TARGET below

lib/vsc/install/shared_setup.py
stdweird: whatever, but i was going to stick to the style of the origin classes

lib/vsc/install/shared_setup.py
stdweird: `setuptools/command/sdist`

lib/vsc/install/shared_setup.py
stdweird: yes, but see the `__main__` in make_release_tree below

lib/vsc/install/shared_setup.py
stdweird: rather not, it's true i don't expect it to be anything else but `shared_setup`, but you never know who will start to subclass this :smile:

lib/vsc/install/shared_setup.py
boegel: yeah, ok, makes sense I guess for this

lib/vsc/install/shared_setup.py
boegel: mention that in the comment, for future reference to where we got this?

lib/vsc/install/shared_setup.py
boegel: ok, fine by me

lib/vsc/install/shared_setup.py
```diff
@@ -156,6 +165,88 @@ def remove_extra_bdist_rpm_files(pkgs=None):
...
+        try:
+            fh = open(dest, 'w')
+            fh.write(source_code)
+            fh.close()
+        except:
+            raise Exception("Failed to write new_shared_setup source to %s" % dest)
+
+    def make_release_tree(self, base_dir, files):
+        log.error("sdist make_release_tree original base_dir %s files %s" % (base_dir, files))
+        log.error("sdist from shared_setup %s current dir %s" % (__file__, os.getcwd()))
```
boegel: do you really want to use `log.error` here?

stdweird: i can't make setuptools show logmessages on any other level :smile:

stdweird:
remarks addressed

stdweird:
@boegel remarks addressed

lib/vsc/install/shared_setup.py
```diff
@@ -156,6 +165,94 @@ def remove_extra_bdist_rpm_files(pkgs=None):
...
+            raise IOError("Failed to write new_shared_setup source to %s (%s)" % (dest, err))
+
+    def make_release_tree(self, base_dir, files):
+        """
+        Create the files in subdir base_dir ready for packaging
+        After the normal make_release_tree ran, we insert shared_setup
+        and modify the to-be-packaged setup.py
+        """
+
+        log.error("sdist make_release_tree original base_dir %s files %s" % (base_dir, files))
```
boegel: include a comment here why you're using `log.error`?

stdweird: done (higher up in the code)

boegel:
OK, going in.

PR #5, opened 2015-10-28
Not zip safe due to vsc/__init__ pkg extend not zip safe

stdweird:


lib/vsc/install/shared_setup.py
```diff
@@ -756,7 +756,6 @@ def action_target(target, setupfn=setup, extra_sdist=[], urltemplate=None):
         'version': VERSION,
         'author': [sdw, ag, jt],
         'maintainer': [sdw, ag, jt],
-        'zip_safe': True,
```
boegel: as mentioned, I would keep this included, set to `False`, and include a comment why

stdweird: it's not an exception o something like that, it's broken everywhere. if people wonder why we ever remeoved it, the commit message looks clear enough.
if we ever manage to work around it, we can probably enable the zif_safe as default

boegel: relevant info: https://github.com/pyinstaller/pyinstaller/issues/1462

possible workaround: https://github.com/NiklasRosenstein/localimport/commit/7c0b9ef96aac54f0b1fc53363e92e9257c78cd9d

PR #11, opened 2015-11-09
Add vsc_release command to show what needs to be done

stdweird:
some smart person might actually make this do stuff instead of just print

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -43,6 +43,7 @@ import setuptools.command.test
 from distutils import log  # also for setuptools
 from distutils.dir_util import remove_tree
 
+from setuptools import Command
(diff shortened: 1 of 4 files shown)


JensTimmerman:
don't we need a clusterbuildrpm and bc step here aswell?

stdweird:
clusterbuildrpm might be a good thing to have, but could be complex.

stdweird:
@JensTimmerman tested with vsc-utils 1.7.3 release

stdweird:
hmmm, github releases can't be used as download urls. they do not have the vsc.install injection like pypi sdist, and also unpack totally wrong (`vsc-utils-1.7.3` tag unpacks in `vsc-utils-vsc-utils-1.7.3`). i removed the release on github

PR #12, opened 2015-11-13
clarify fixing common test issues

stdweird:


stdweird:
@JensTimmerman last commit needed for corretc rpm building

lib/vsc/install/shared_setup.py
```diff
@@ -144,13 +144,13 @@ def _log(self, level, msg, args):
...
-VERSION = '0.9.8'
+VERSION = '0.9.9'
 
 log.info('This is (based on) vsc.install.shared_setup %s' % VERSION)
 
-# list of non-vsc packages that need python- prefix for correct rpm dependencies
+# list of non-vsc packages that do not need python- prefix for correct rpm dependencies
 # vsc packages should be handled with clusterbuildrpm
-PREFIX_PYTHON_BDIST_RPM = ('setuptools',)
+NO_PREFIX_PYTHON_BDIST_RPM = ()
```
JensTimmerman: this looks confusing, you might want to change `()` to `tuple()` so it's clear what this is supposed to be?

JensTimmerman: actually, can't this be a list? So you can maybe extend it later on, or is this exactly what you don't want?
in that case, maybe document that?

JensTimmerman:
jenkins: test this please

PR #14, opened 2015-11-23
Add more python prefix handling options

stdweird:


JensTimmerman:
needs tests

JensTimmerman:
hmm, weird error, let's try again, jenkins: test this please

lib/vsc/install/shared_setup.py
```diff
@@ -144,13 +144,19 @@ def _log(self, level, msg, args):
...
 
 # list of non-vsc packages that do not need python- prefix for correct rpm dependencies
 # vsc packages should be handled with clusterbuildrpm
-NO_PREFIX_PYTHON_BDIST_RPM = []
+# dependencies starting with python- are also not re-prefixed
+NO_PREFIX_PYTHON_BDIST_RPM = ['pbs_python']
+
+# Hardcode map of python dependency prefix to their rpm python- flavour prefix
+PYTHON_BDIST_RPM_PREFIX_MAP = {
+    'pycrypto': 'python-crypto',
```
itkovian: Maybe we should have a little intro on this, so people know they also have to change things here?

lib/vsc/install/shared_setup.py
```diff
@@ -664,43 +724,37 @@ def setup_sys_path(self):
...
-                # only actual modules
-                reload_modules.append(name)
-            del(sys.modules[name])
+        # reload the loaded modules with new __path__
+        for module in reloaded_modules:
+            try:
+                __import__(module)
+                log.debug('Imported module %s' % module)
+            except ImportError as e:
+                log.info('Failed to reload module %s, continuing: %s' % (module, e))
```
JensTimmerman: maybe we should actually fail for now, so we can really pick up when something goes wrong.

lib/vsc/install/shared_setup.py
stdweird: make an issue with you want to see documented

stdweird:
@JensTimmerman done

lib/vsc/install/shared_setup.py
```diff
@@ -664,43 +724,37 @@ def setup_sys_path(self):
...
-                # only actual modules
-                reload_modules.append(name)
-            del(sys.modules[name])
+        # reload the loaded modules with new __path__
+        for module in reloaded_modules:
+            try:
+                __import__(module)
+                log.debug('Imported module %s' % module)
+            except ImportError as e:
+                raise ImportError("Failed to reload module %s: %s" % (module, e))
```
ehiggs: Given how `__import__('does_not_exist')` will raise `ImportError(`No module named %s' e)` do we really need to try, except and reraise with a new message? At this point, I think we can elide the noise and just call import and let the underlying machinery handle the error.

lib/vsc/install/shared_setup.py
```diff
@@ -966,21 +1019,30 @@ def sanitize(name):
...
+
+            # more sensible map
+            p_p = (not ([x for x in NO_PREFIX_PYTHON_BDIST_RPM if name.startswith(x)] or name.startswith('python-'))
+                   or name.startswith('vsc'))
+            if p_p:
+                name = 'python-%s' % name
         return name
-
-    return ",".join([sanitize(r) for r in name])
+    else:
```
ehiggs: We are at the bottom of the function; no need for `else`.

lib/vsc/install/shared_setup.py
stdweird: the original error only shows the module, not the package

lib/vsc/install/shared_setup.py
stdweird: it's not some sort of default/fallback, it's really something `else`

lib/vsc/install/shared_setup.py
```diff
@@ -144,13 +144,19 @@ def _log(self, level, msg, args):
 
 RELOAD_VSC_MODS = False
 
-VERSION = '0.9.10'
+VERSION = '0.9.12'
```
ehiggs: This is adding features, so it's a minor version; not a bug fix bump. As it's <0.* I guess all bets are off on the actual versioning scheme.

lib/vsc/install/shared_setup.py
JensTimmerman: it shouldn't be needed per se, but I like how it makes the recursion condition more explicit.
Less confusion in the future when reading is ok to stay for me.

lib/vsc/install/shared_setup.py
JensTimmerman: I sort of agree with Ewan here, 
you can remove the whole try except block, and have less code to worry about.
```
>>> __import__('foo.bar')
  File "<stdin>", line 1, in <module>
ImportError: No module named foo.bar
>>> 
```

stdweird: try with .`vsc.foo.bar`
```
>>> __import__('vsc.foo.bar')
  File "<stdin>", line 1, in <module>
ImportError: No module named foo.bar
```

test/shared_setup.py
```diff
@@ -15,10 +16,52 @@ def test_get_name_url(self):
...
                              msg='determined name and url from %s file' % fn)
 
         res.pop('download_url')
         fn = 'git_config'
-        self.assertEqual(get_name_url(os.path.join(REPO_TEST_DIR, 'setup', fn), version='0.1.2', license_name='LGPLv2+'), res,
+        self.assertEqual(get_name_url(os.path.join(shared_setup.REPO_TEST_DIR, 'setup', fn), version='0.1.2', license_name='LGPLv2+'), res,
                          msg='determined name and url from %s file with license' % fn)
+
+    def test_sanitize(self):
+        """Test sanitize function"""
```
ehiggs: Some python packages are named with a `python-` prefix already. e.g. [`python-jenkins`](https://pypi.python.org/pypi/python-jenkins/0.4.11), [`python-etcd`](https://pypi.python.org/pypi/python-etcd/0.4.2). Please put in a test with these to make sure we do or don't sanitize to `python-python-jenkins` (I don't know which is better/worse, but it would be good to have it documented and tested so we can catch unexpected changes).

test/shared_setup.py
```diff
@@ -15,10 +16,52 @@ def test_get_name_url(self):
...
+        os.environ['VSC_RPM_PYTHON']='1'
+        self.assertEqual(sanitize('anything >= 5'), 'python-anything >= 5',
+                         msg='all packages are prefixed with VSC_RPM_PYTHON set')
+
+        self.assertEqual(sanitize('vsc-xyz >= 10'), 'python-vsc-xyz >= 10',
+                         msg='vsc packages are prefixed with VSC_RPM_PYTHON set')
+
+        self.assertEqual(sanitize('python-ldap == 11'), 'python-ldap == 11',
```
ehiggs: This is a good test to add in. Can we also add `python_*` such as [`python_webdav`](https://pypi.python.org/pypi/python_webdav/0.4]

PR #19, opened 2016-01-15
prospector is optional, except on jenkins

stdweird:
rpm creation is a nightmare

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -1297,10 +1297,14 @@ if __name__ == '__main__':
         'setuptools',
     ]
     if sys.version_info < (2, 7):
-        # py26 support dropped in 0.8, and theo old versions don't detect enough
-        log.info('no prospector support')
+        # py26 support dropped in 0.8, and the old versions don't detect enough
+        log.info('no prospector support in py26 (or older)')
     else:
-        install_requires.append('prospector >= 0.11.7')
+        # For now, not enforcing propspector.
+        # It would complicate rpm creation too much.
+        # Should be re-added when the testing code is moved to vsc-testing
+        log.info('not enforcing prospector support. run "easy_install propspector" yourself')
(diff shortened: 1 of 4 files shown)


lib/vsc/install/testing.py
```diff
@@ -284,7 +286,15 @@ def test_prospector(self):
...
         if not HAS_PROTECTOR:
-            log.info('No protector tests are ran')
+            if sys.version_info < (2, 6):
+                log.info('No protector tests are ran on py26 or older.')
+            else:
+                log.info('No protector tests are ran, install prospector manually first')
+
+                # This is fatal on jenkins/...
+                if 'JENKINS_URL' in os.environ:
+                    self.assertTrue(False, 'prospecotr must be installed in jenkins environment')
```
JensTimmerman: typo: prospecotr -> prospector

lib/vsc/install/shared_setup.py
```diff
@@ -1297,10 +1297,14 @@ def action_target(target, setupfn=setup, extra_sdist=[], urltemplate=None):
...
-        log.info('no prospector support')
+        # py26 support dropped in 0.8, and the old versions don't detect enough
+        log.info('no prospector support in py26 (or older)')
     else:
-        install_requires.append('prospector >= 0.11.7')
+        # For now, not enforcing propspector.
+        # It would complicate rpm creation too much.
+        # Should be re-added when the testing code is moved to vsc-testing
+        log.info('not enforcing prospector support. run "easy_install propspector" yourself')
+        #install_requires.append('prospector >= 0.11.7')
```
JensTimmerman: use package extra's to make this easy?
easy_install package[extras] should then install all optional extra's
 http://pythonhosted.org/setuptools/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

stdweird: ok, will try that and see what happens when rpms get build.

Issue #21, opened 2016-02-03
shared setup should specify that `vsc` in a namespace package


boegel:
With the vsc-base/vsc-install split, EasyBuild users are running into warnings like:

```
/usr/local/easybuild/2.6.0/software/EasyBuild/2.6.0/lib/python2.6/site-packages/vsc_base-2.4.16-py2.6.egg/vsc/__init__.py:29:
UserWarning: Module vsc was already imported from
/usr/local/easybuild/2.6.0/software/EasyBuild/2.6.0/lib/python2.6/site-packages/vsc_base-2.4.16-py2.6.egg/vsc/__init__.pyc,
but /usr/local/easybuild/2.6.0/software/EasyBuild/2.6.0/lib/python2.6/site-packages/vsc_install-0.9.12-py2.6.egg is being added to sys.path

  import pkg_resources
```

(cfr. https://github.com/hpcugent/easybuild-framework/issues/1588)

It seems like this can be solved by specifying that `vsc` is a namespace packages, see https://pythonhosted.org/setuptools/setuptools.html#namespace-packages.

stdweird:
currenlty, `vsc` is not a pure namespace in vsc-install, it holds `vsc.fancylogger` for backward compatible reasons. but it can be a good default.
however, the magic is in the `__init__`, which should already be correct. can you ask the content of the vsc/__init__ in both vsc-install egg and vsc-base?

boegel:
I'm not saying the ~~magic~~ content in the `__init__.py` is wrong, not at all.

Only one `__init__.py` will ever be executed (the others will be scanned, but their contents will not be executed). That's the case already.

The warnings happen because Python (or setuptools, whatever) notices that the `vsc/__init__.py` files are present in multiple different places on the filesystem (i.e. in different `.eggs` subdirectories).

If we register `vsc` as a namespace package in the `setup.py`, we're just telling Python that's exactly what we want to do, and it will stop warning about this.

Note that we get basically the same warning in the case where `vsc-base` is available both in the EasyBuild installation prefix *and* on the OS (there it *is* wrong, and the warning is legitimate)
We'll need to check whether the warning is still going to occur in that case. We would want that, I think...

stdweird:
well, the magic content in the eb `__init__` is wrong. i also do not understand the original error message, as eb doesn't use `pkg_resources`.

the warnings can't come only from your exlpanation, as i've never seen them, and i have quite a few more packages extending `vsc` namespace installed.

[...]

Issue #16, opened 2015-11-27
sdist upload to PyPI fails


boegel:
I ran into this while trying to upload vsc-install 0.9.12 to PyPI:

```
$ python setup.py sdist upload
INFO: This is (based on) vsc.install.shared_setup 0.9.12
INFO: run_tests from base dir /Users/kehoste/work/vsc-install (using executable /Users/kehoste/work/vsc-install/setup.py)
WARN: cleanup lib/vsc_install.egg-info
INFO: generated list: ['vsc.install', 'vsc']
INFO: generated packages list: ['vsc.install', 'vsc']
INFO: makesetupcfg set to True, (re)creating setup.cfg
INFO: found license /Users/kehoste/work/vsc-install/LICENSE with md5sum 5f30f0716dfdd0d91eb439ebec522ec2
INFO: Found license name LGPLv2+ and classifier License :: OSI Approved :: GNU Lesser General Public License v2 or later (LGPLv2+)
INFO: setting license LGPLv2+
INFO: No name defined, trying to determine it
INFO: found match url git@github.com:hpcugent/vsc-install in /Users/kehoste/work/vsc-install/.git/config
INFO: found match name vsc-install in /Users/kehoste/work/vsc-install/.git/config
INFO: Removing None download_url
INFO: get_name_url returns {'url': 'https://github.com/hpcugent/vsc-install', 'name': 'vsc-install'}
INFO: using long_description vsc-install provides shared setuptools functions and classes for python libraries developed by UGent's HPC group
INFO: generated list: []
INFO: generated scripts list: []
INFO: running sdist
INFO: running egg_info
INFO: creating lib/vsc_install.egg-info
INFO: writing requirements to lib/vsc_install.egg-info/requires.txt
INFO: writing lib/vsc_install.egg-info/PKG-INFO
INFO: writing top-level names to lib/vsc_install.egg-info/top_level.txt
INFO: writing dependency_links to lib/vsc_install.egg-info/dependency_links.txt
INFO: writing manifest file 'lib/vsc_install.egg-info/SOURCES.txt'
INFO: reading manifest file 'lib/vsc_install.egg-info/SOURCES.txt'
INFO: writing manifest file 'lib/vsc_install.egg-info/SOURCES.txt'
INFO: looking for extra dist files
WARN: warning: vsc_sdist: standard file not found: should have one of README, README.rst, README.txt

INFO: running check
[...]

PR #22, opened 2016-02-06
register 'vsc' as namespace package

boegel:
fix for #21

I'd love to figure out #16 in this PR, but not sure yet which setuptools to require...

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -1049,6 +1049,7 @@ SHARED_TARGET = {
         "vsc_release": vsc_release,
     },
     'command_packages': ['vsc.install.shared_setup', NEW_SHARED_SETUP, 'setuptools.command', 'distutils.command'],
+    'namespace_packages': ['vsc'],
     'download_url': '',
     'package_dir': {'': DEFAULT_LIB_DIR},
     'setup_requires' : ['setuptools', 'vsc-install >= %s' % VERSION],
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -1049,7 +1049,6 @@ SHARED_TARGET = {
         "vsc_release": vsc_release,
     },
     'command_packages': ['vsc.install.shared_setup', NEW_SHARED_SETUP, 'setuptools.command', 'distutils.command'],
-    'namespace_packages': ['vsc'],
     'download_url': '',
     'package_dir': {'': DEFAULT_LIB_DIR},
     'setup_requires' : ['setuptools', 'vsc-install >= %s' % VERSION],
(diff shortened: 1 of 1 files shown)


lib/vsc/install/shared_setup.py
```diff
@@ -1357,6 +1357,7 @@ def action_target(target, setupfn=setup, extra_sdist=[], urltemplate=None):
             'setuptools'
         ],
         'excluded_pkgs_rpm': [], # vsc-install ships vsc package (the vsc package is removed by default)
+        'namespace_packages': ['vsc'],
```
stdweird: as said before, for `vsc-install`, `vsc` is not a pure namespace; but it should be set as default for all other vsc packages

boegel: well, I didn't understand your remark w.r.t. 'pure' namespace; if you spread `vsc` across multiple locations, you need this, and this applies to `vsc-install` too, since it provides stuff accessible through `from vsc.install import ...`

How would I set this as the default for all other vsc packages?

Anyway, I don't think that matters much, we need this for `vsc-install` itself too imho.

stdweird: my understanding of a namespace package is that the namespace itself is empty. so as soon as we remove `vsc.fancylogger`, this is ok.
@JensTimmerman is this a good time to remove `vsc.fancylogger`? we should be able to check logstash for any leftover usage.

boegel: From http://peak.telecommunity.com/DevCenter/setuptools#namespace-packages:

```
When you declare a package to be a namespace package, it means that the package has no meaningful contents
in its __init__.py, and that it is merely a container for modules and subpackages.
```

So, both modules `fancylogger` (which is just a stub, but fine) and subpackages are allowed in namespaces.

However, I do support getting rid of `vsc.fancylogger`.

stdweird:
this does not set the default for teh vsc modules
and do we want to get rid of `vsc.fancylogger` or not?

JensTimmerman:
yes, vsc.fancylogger should go away

boegel:
@stdweird: any pointers on how to set the default for the `vsc` modules?

stdweird:
look for the `SHARED_TARGET`?

boegel:
@stdweird: done

stdweird:
no, it's not. i'll fix this tomorrow...

boegel:
@stdweird: just tell me what I did wrong?

boegel:
@stdweird: remarks fixed, please rereview?

Not sure if we should also chuck `vsc.fancylogger` in this PR...

Maybe that should be done in a separate PR, together with a version bump to 1.0?

boegel:
Jenkins: test this please

boegel:
Jenkins: ok to test

boegel:
Jenkins: test this please

PR #23, opened 2016-02-08
fix formatting issues in README

boegel:
Fix formatting issues that prevent proper rendering on PyPI

Checks done using readme-renderer Python package, using python setup.py check -rs, cfr. https://gist.github.com/kennethreitz/8378ba05d1a50bea75da

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,14 +44,13 @@ python setup.py test --help
 
 In case following error occurs, it means there is a test module `XYZ` that cannot be imported.
 
-```
-  ...
-  File "setup.py", line 499, in loadTestsFromModule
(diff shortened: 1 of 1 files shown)


stdweird:
This needs to be added as a check so all repos are tested

stdweird:
@boegel what renderer version did you use? it still gives errors, see failing test output from https://github.com/hpcugent/vsc-install/pull/28 (https://jenkins1.ugent.be/job/vsc-install-prbuilder/PYTHON=CPython-2.7/45/console)

boegel:
I installed the latest version from PyPI (0.7.0)

PR #25, opened 2016-02-08
Handle the prospector pyroma hack

stdweird:


stdweird:
@JensTimmerman the fix in #24 wasnot enough. i also opened a PR with prospector to solve it properly

PR #26, opened 2016-02-09
Address pylint monkeypatch help formatter

stdweird:


stdweird:
@JensTimmerman this should solve the `__init__` errors in vsc-base tests

PR #31, opened 2016-03-04
specify minimal required setuptools version

boegel:


boegel:
cfr. https://github.com/hpcugent/vsc-install/issues/16

boegel:
Hmm, this is what we have on SL6 by default...

```
r-bash-4.1$ rpm -qa | grep setuptools
python-setuptools-0.6.10-3.el6.noarch

-bash-4.1$ python
Python 2.6.6 (r266:84292, Jul 22 2015, 16:47:47) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-16)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import setuptools
>>> setuptools.__version__
'0.6' 
```

Thoughts @stdweird?

boegel:
Even CentOS7 is quite old...

```
$ python -c "import setuptools; print setuptools.__version__"
0.9.8
```

boegel:
OK, now also specifying that 0.6 is minimal runtime dep, while 2.0 is a conservative minimal dep for running `setup.py` (cfr. #16)...

stdweird:
building rpms fails, please actually this instead of assuming how it works
```
INFO: generated scripts list: []
  File "setup.py", line 1366, in <module>
    action_target(PACKAGE)
  File "setup.py", line 1335, in action_target
    setupfn(**x)
  File "/usr/lib64/python2.7/distutils/core.py", line 112, in setup
    _setup_distribution = dist = klass(attrs)
  File "/usr/lib/python2.7/site-packages/setuptools/dist.py", line 265, in __init__
    self.fetch_build_eggs(attrs.pop('setup_requires'))
  File "/usr/lib/python2.7/site-packages/setuptools/dist.py", line 289, in fetch_build_eggs
    parse_requirements(requires), installer=self.fetch_build_egg
  File "/usr/lib/python2.7/site-packages/pkg_resources.py", line 630, in resolve
    raise VersionConflict(dist,req) # XXX put more info here
pkg_resources.VersionConflict: (setuptools 0.9.8 (/usr/lib/python2.7/site-packages), Requirement.parse('setuptools>=2.0'))
```

boegel:
@stdweird I didn't make any assumptions, I actually mentioned that the setuptools version on both SL6 and CentOS7 was too old compared to what I have in here...

setuptools 2.0 is only required to run `sdist upload` without problems (at least on my system), but not for installing or building the RPM, so I guess we can lower the version requirement?

boegel:
@stdweird ping on my question above?

JensTimmerman:
yes, better lower it.

stdweird:
so you are going to lower the version requirement to somethig less than 2.0 yet make it at least 2.0 in order to fix the original issue? can you update the PR so it is clear what you are trying to do and that we can still build rpms?

boegel:
@stdweird well, I'm open to suggestions...

For running `setup.py bdist_rpm`, 0.6 is enough (since we've been building RPMs on SL6 without a problem), for running `setup.py sdist upload`, it appears 0.6 is not sufficient (2.0 is, there may be a slightly lower version that is also sufficient).

There's no separate version requirement I can specify for `sdist upload` though, at least not in a clean way...

boegel:
apparently this turns out to be pretty much useless, so closing

PR #32, opened 2016-03-13
Clarify missing url exception

stdweird:


boegel:
No test changes to check for this?

stdweird:
like? this is pure informational, the keyerror would occur on the next line, but with a message that isn't helpful

boegel:
fine

PR #34, opened 2016-05-03
cleaned up:

JensTimmerman:
- pep8 and pylint are now complaining less
 - added a few very interesting errors as failures in prospector: dangerous-default-value and no-value-for-parameter
 - moved path detecting code in shared_setup into it's own namespace, so it is rerun when we run shared_setup inside shared_setup (e.g. when installing a package trought python setup.py install, and it's dependencies are also importing the system installed shared_setup, instead of the only current working solution that was installing from pypi with a patched in copy of vsc_install

### DIFF
--- a/lib/vsc/install/commontest.py
+++ b/lib/vsc/install/commontest.py
@@ -41,8 +41,7 @@ import re
 import sys
 
 from distutils import log
-from vsc.install.shared_setup import generate_packages, generate_scripts, generate_modules, \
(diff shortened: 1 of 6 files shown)


JensTimmerman:
this ended up all in one big commit, and one pr, 
mainly because my local .git folder got corrupted and I had to restore it from a fresh .git folder, loosing all my history (but luckily not my changes) :-(

PR #40, opened 2016-05-17
fix broken link to VSC website

boegel:


stdweird:
@boegel version bump?

boegel:
OK, here it makes sense I guess, since it contains code that checks the headers...

Done!

PR #42, opened 2016-05-19
Fix rpm exclude pkgs

JensTimmerman:
the excluded pgkgs ware never picked up when we started to cache the package_files, 
so be sure to regenerate the package files with the excluded packages as parameter, once we know it.

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -1375,9 +1375,11 @@ if __name__ == '__main__':
         'version': VERSION,
         'author': [sdw, ag, jt],
         'maintainer': [sdw, ag, jt],
-        'excluded_pkgs_rpm': [],  # vsc-install ships vsc package (the vsc package is removed by default)
         'install_requires': install_requires,
-        'setup_requires': 'setuptools',
+        'setup_requires': [
+            'setuptools',
+        ],
+        'excluded_pkgs_rpm': [],  # vsc-install ships vsc package (the vsc package is removed by default)
     }
 
     action_target(PACKAGE)
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -1358,6 +1358,9 @@ if __name__ == '__main__':
     """
     This main is the setup.py for vsc-install
     """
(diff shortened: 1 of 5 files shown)


lib/vsc/install/shared_setup.py
```diff
@@ -1353,9 +1358,6 @@ def action_target(package, *args, **kwargs):
     """
     This main is the setup.py for vsc-install
     """
-    install_requires = [
```
stdweird: i wouldn't remove it, as we might start to add minimal version requirements, and then this becomes useful. otoh, if we would need them, we could readd them. but this basically means: we support whatever el6 ships.

boegel: cfr. #31

stdweird: except that that PR doesn't actually adds any restrictions. it'll be quite a challenge to get < 0.6 :smile:

JensTimmerman: the problem is, currently clusterbuildrpm is doing the translation from python- prefixed to not python- prefixed (I guess?), so it works on our buildhosts with clusterbuildrpm, but it doesn't work locally on my machine, because when I run `python setup.py bdist_rpm` I get
```
error: Failed build dependencies:
	setuptools is needed by vsc-install-0.10.6-1.noarch
error: command 'rpmbuild' failed with exit status 1
```

so we should maybe move these translation thingies from clusterbuildrpm to vsc-install to fix this?

stdweird: there's a vsc environment variable in vsc-install to control the renaming of the dependencies, but clusterbuildrpm does the actually renaming of the rpm using rpmrebuild. i have not looked into patching bdist_rpm to produce the "correct" python- named rpms itself.

JensTimmerman: right, I forgot about `export VSC_RPM_PYTHON=1` works like a charm :-)

stdweird:
@JensTimmerman code looks ok. but are you finished?

JensTimmerman:
yes, this branch of vsc-install is already installed on regice and regirock, where it works now.

PR #44, opened 2016-06-21
Import message

itkovian:


lib/vsc/install/commontest.py
```diff
@@ -122,7 +122,7 @@ def _import(self, pkg):
         except ImportError as e:
             log.warn('Import of %s failed with %s' % (pkg, e))
 
-        self.assertTrue(pkg in sys.modules, msg='import %s was success' % pkg)
+        self.assertTrue(pkg in sys.modules)
```
boegel: wait, won't this hide the value of `pkg`, and result in an even less informative error message?

stdweird: hmm, good point. i'll revert this, also since the error message is the one that the warn should log above.

PR #45, opened 2016-06-22
Revert pr 44

stdweird:


lib/vsc/install/commontest.py
```diff
@@ -122,7 +122,7 @@ def _import(self, pkg):
         except ImportError as e:
             log.warn('Import of %s failed with %s' % (pkg, e))
 
-        self.assertTrue(pkg in sys.modules)
+        self.assertTrue(pkg in sys.modules, msg='import %s was success' % pkg)
```
boegel: how about doing an `assertTrue(False, ...)` in the `except` above, where you can include the error in the message?

stdweird:
@boegel updated

lib/vsc/install/commontest.py
JensTimmerman: why not just let the error get raised ?

stdweird: ?

PR #46, opened 2016-06-30
Add _fvs primitive subclassing support

stdweird:


stdweird:
@JensTimmerman support subclassing sort-of

JensTimmerman:
I've looked into how an inner class can get access to the outer class, see http://stackoverflow.com/questions/2278426/inner-classes-how-can-i-get-the-outer-class-object-at-construction-time/2278496#2278496

this way we don't always need to create an instance of vsc_setup, but can use `self.outer`

and inheratance would allow you to do 
```
class my_vsc_setup(vsc_setup):
   #redefine some inner class functionality here
   @InnerClassDescriptor
   class vsc_sdist(object):
        self.outer.some_function()
         pass     
```

maybe that is a bit less hacky then this? ( I should have looked at this in the first place :S)

PR #47, opened 2016-07-07
More lint

JensTimmerman:


lib/vsc/install/commontest.py
```diff
@@ -86,21 +86,27 @@ class CommonTest(TestCase):
...
+        #'wrong-import-position',  # not sure about this, these usually have a good reason
     ]
     PROSPECTOR_WHITELIST = [
         'undefined',
         'no-value-for-parameter',
         'dangerous-default-value',
         'redefined-builtin',
-        #'bare-except',
+        'bare-except',
+        'E713',
```
stdweird: add small comment that is a bit descriptive

lib/vsc/install/commontest.py
```diff
@@ -86,21 +86,27 @@ class CommonTest(TestCase):
...
+        'arguments-differ',
+        'unused-argument',
+        'unused-variable',
+        'reimported',
+        'F811',
```
stdweird: same here, add 1 or 2 words explaining/reminding people what this is about

stdweird:
so how bad is this going to be?

JensTimmerman:
it is still pretty limited, everything points to either errors/issues that need to be addressed, or things that are against our coding guidelines. nothing pedantic yet, that will come later ;)

stdweird:
i'll merge this if you send a small mail to hpc-admin explaining what to expect typically :grin:

JensTimmerman:
mail sent

JensTimmerman:
overview of how to handle these things:
```

bare-except
============
try:
   # something
except:

This is bad, because this except will also catch sys.exit() or Keyboardinterupts, something you 
typically do not want, if you catch these the program will be in a weird state and then continue on, 
whilst the person who just pressed ctrl+c is wondering what is going on and why it is not stopping.

so at the very least make this
except Exception (which doesn't catch sys.exit and KeyboardInterupt)
and it would be appreciated if you could actually figure out what exceptions to expect and only catch those
and let your program crash if something you did not intend happens 
because it helps developers catch weird errors on their side better.

if you do something like
try:
    open(int(somestring)).write('important data')
except Exception:
    pass # if somestring is not an integer, we didn't need to write anyway, but otherwise we do

because you know sometimes this string does not contain an integer, so the int() call can fail
you should really only catch ValueError, because this will also fail when your disk is full, or you don't have permissions 
or xxx other reasons, and the important data will not be written out and nobody will notice anything!



if not 'a' in somelist -> if 'a' not in somelist
=================================================

this isn't that big of a deal, but if everyone is consistent it's less likely to introduce bugs when a not is added or removed where it didn't need to.
Also helps code review, not in reads better, like english.


arguments-differ
=================

this will give you errors if you override a function of a superclass but don't use the same amount of arguments, 
using less will surely give you errors, so the linter catches this for you now

unused-argument
===============
if you have a function definition witch accepts an argument that is never used in the function body this will now give an error. 
clean up your function definition, or fix the error where you actually do need to take this argument into account

unused-variable
================
defining a variable and then not using it anymore smells bad, why did you do that?
sometimes you do things like

out, exit_code = run_command(something) 
but you are not interested in the out, only in the exit code, 
in this case, write 
_, exit_code = run_command(something) 

using _ as a variable name lets pylint and other readers know you do not intend to use that output in the first place.


reimported
===========
when you re import a name somewhere else, 
usually this is just an import to much, or 2 imports with the same name, pay attention.

import six
from django import six
=>

import six
from django import six as django_six


redefinition of unused name
============================
this usually also points to something you did not expect

from vsc.accountpageclient import VscGroup
<snip>

class VscGroup(object):
    pass

=> do you need the import? use import as 
did you mean to use the same name? ...




if in any of these cases you think: yes, I really needed to do this, I'm monkeypatching things, I'm adding extra functionality that does indeed have an extra(default) paramenter, etc, etc
you can let pylint know to ignore this error in this one specific block of code
by adding e.g.the comment

class Something(object):
    def dosomething(self, some, thing):
        # do something

class MyFancyThing(SomeThing):
    # pylint: disable=arguments-differ
    def dosomething(self, some, thing, fancy=None):
         # do soemthing fancy

```

boegel:
@JensTimmerman add the latter to a README or something in this repo?

JensTimmerman:
@boegel  https://github.com/hpcugent/vsc-install/pull/49

PR #51, opened 2016-08-29
Fix fix headers exception handling

stdweird:


stdweird:
@JensTimmerman

lib/vsc/install/headers.py
```diff
@@ -314,7 +314,7 @@ def check_header(filename, script=False, write=False):
     args = sys.argv[1:]
     try:
         is_script = int(args[-1]) == 1
-    except IndexError, ValueError:
+    except (IndexError, ValueError):
```
JensTimmerman: let's whitelist the pylint warning: `unpacking-in-except`, (E1603, W0712) so this is caught in the future.

PR #28, opened 2016-02-09
Add readme_renderer support

stdweird:
fixes #27

PR #54, opened 2016-09-01
Add readme_renderer support

JensTimmerman:
fixes remarks in #28

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ Description
 vsc-install provides shared setuptools functions and classes for python libraries developed by UGent's HPC group
 
 Common pitfalls
-=========
+===============
 bdist_rpm will fail if your install_requires = 'setuptools' because it will fail to find a setuptools rpm.
 ```
 export VSC_RPM_PYTHON=1
@@ -54,10 +54,10 @@ In case following error occurs, it means there is a test module `XYZ` that canno
 
 ```txt
 File "setup.py", line 499, in loadTestsFromModule
-    testsuites = ScanningLoader.loadTestsFromModule(self, module)
+testsuites = ScanningLoader.loadTestsFromModule(self, module)
 File "build/bdist.linux-x86_64/egg/setuptools/command/test.py", line 37, in loadTestsFromModule
 File "/usr/lib64/python2.7/unittest/loader.py", line 100, in loadTestsFromName
-    parent, obj = obj, getattr(obj, part)
+parent, obj = obj, getattr(obj, part)
 AttributeError: 'module' object has no attribute 'XYZ'
 ```
 
@@ -115,7 +115,7 @@ bare-except
 -----------
 ```python
 try:
-   # something
+# something
 except:
 ```
 This is bad, because this except will also catch sys.exit() or Keyboardinterupts, something you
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -54,10 +54,10 @@ In case following error occurs, it means there is a test module `XYZ` that canno
 
 ```txt
 File "setup.py", line 499, in loadTestsFromModule
-    testsuites = ScanningLoader.loadTestsFromModule(self, module)
+testsuites = ScanningLoader.loadTestsFromModule(self, module)
 File "build/bdist.linux-x86_64/egg/setuptools/command/test.py", line 37, in loadTestsFromModule
 File "/usr/lib64/python2.7/unittest/loader.py", line 100, in loadTestsFromName
-    parent, obj = obj, getattr(obj, part)
+parent, obj = obj, getattr(obj, part)
```
wpoely86: why is the indentation being removed?

lib/vsc/install/commontest.py
```diff
@@ -212,3 +215,27 @@ def test_prospector(self):
...
 
         self.assertFalse(failures, "prospector failures: %s" % pprint.pformat(failures))
+
+    def test_readme_renderer(self):
+        """
+        The README has to render properly.
+        """
+
+        # inspired by the check -rs command readme_renderer.integration.distutils
+
```
wpoely86: less empty lines?

lib/vsc/install/commontest.py
```diff
@@ -212,3 +215,27 @@ def test_prospector(self):
         optparse.HelpFormatter.expand_default = orig_expand_default
 
         self.assertFalse(failures, "prospector failures: %s" % pprint.pformat(failures))
+
+    def test_readme_renderer(self):
+        """
+        The README has to render properly.
+        """
+
+        # inspired by the check -rs command readme_renderer.integration.distutils
```
wpoely86: why not put this in the doc string?

lib/vsc/install/commontest.py
```diff
@@ -212,3 +215,27 @@ def test_prospector(self):
...
+
+        # existence of README is checked in shared_setup
```
wpoely86: nevertheless, shouldn't this be put in a try except clause? It's not because it exists that you can read it.

README.md
JensTimmerman: because it's not valid rst otherwise, which is what this pr is supposed to check for

wpoely86: but it's a code block, it isn't rst to begin with.

boegel:
all of this is pretty useless unless we switch to `README.rst` in all `vsc-*` repos, because both `readme-renderer` and PyPI only support RST...

README.md
JensTimmerman: The purpose of the original pr that I am fixing here ( #28 ) was to check all README files to render as proper rst

that's all I'm doing here.
I guess we should decide if it's the way we want to go.
currently this file now renders ok as markdown, and doesn't fail on rst,

but if we want to go ahead and merge this, all readme's must become rst.

JensTimmerman:
agreed, this is pretty useless, I'm not for changing all md to rst just because pypi can't handle it yet.
they'll add support one day. so, let's not fix?
(but it does fix the errors in #28 )

lib/vsc/install/shared_setup.py
```diff
@@ -1122,7 +1122,7 @@ def run(self):
         'url': '',
         'dependency_links': [],
         'install_requires': [],
-        'tests_require': ['nose', 'mock'],
+        'tests_require': ['nose', 'mock', 'readme_renderer'],
```
stdweird: should be appended like prospector

README.md
stdweird: try to only run the renderer tests if threadme is a a `.rst`

PR #55, opened 2016-09-12
fix for TypeError: dist must be a Distribution instance

JensTimmerman:


lib/vsc/install/shared_setup.py
```diff
@@ -1395,12 +1394,16 @@ def prepare_rpm(self, target):
...
         self.build_setup_cfg_for_bdist_rpm(target)
 
-    def action_target(self, target, setupfn=setup, extra_sdist=None, urltemplate=None):
+    def action_target(self, target, setupfn=None, extra_sdist=None, urltemplate=None):
         """
         Additional target attributes
         makesetupcfg: boolean, default True, to generate the setup.cfg (set to False if a manual setup.cfg is provided)
         provides: list of rpm provides for setup.cfg
         """
+        if not setupfn:
```
stdweird: `is not None`

lib/vsc/install/shared_setup.py
```diff
@@ -1395,12 +1394,16 @@ def prepare_rpm(self, target):
...
+            #  late import, so were don't accidentally use the distutils setup
```
stdweird: can you paste the a ref to the issue here?

PR #56, opened 2016-09-26
add dependency links for private vsc-* dependencies

JensTimmerman:


JensTimmerman:
this started working as we like since setuptools 27.0, so update your setuptools if you now can't find dependencies in tests. (no need to any longer suply dependency links in setup.py for our vsc-* projects thare are hosted on github)

lib/vsc/install/shared_setup.py
```diff
@@ -1328,7 +1328,30 @@ def parse_target(self, target, urltemplate=None):
...
+            ]
+        else:
+            urls = [('github.com', 'git+https://')]
+        for dependency in set(new_target['install_requires'] + new_target['setup_requires'] +
+            new_target['tests_require']):
+            if dependency.startswith('vsc'):
+                dep = dependency.split(' ')[0]
+                depversion = ''
+                for comp in ['=', '<']:
+					try:
```
stdweird: moar whitespace

stdweird:
@JensTimmerman version bump pls

boegel:
@JensTimmerman we don't have a style check for (well, against) tabs?

JensTimmerman:
@boegel prospector tests all sorts of things, we currently don't have everything turned on yet because it would probably be to annoying to have to rework all the code, however, maybe we should, because every check has a good reason, and as we turn on more and more over time it becomes annoying to update all code a little bit every time...
anyway, turned on check for mixing tabs and spaces.

PR #58, opened 2016-10-11
fix attribute error for uknown attributes in the vsc_setup_klass

JensTimmerman:
fixes https://github.com/hpcugent/vsc-base/issues/241

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -146,7 +146,7 @@ URL_GHUGENT_HPCUGENT = 'https://github.ugent.be/hpcugent/%(name)s'
 
 RELOAD_VSC_MODS = False
 
-VERSION = '0.10.16'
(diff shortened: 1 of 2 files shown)


lib/vsc/install/shared_setup.py
```diff
@@ -1227,7 +1227,10 @@ def parse_target(self, target, urltemplate=None):
         # update the cmdclass with ones from vsc_setup_klass
         # cannot do this in one go, whne SHARED_TARGET is defined, vsc_setup doesn't exist yet
```
boegel: @JensTimmerman fix this typo `whne`?

boegel:
fix verified, lgtm

lib/vsc/install/shared_setup.py
boegel: ah, not fixed yet...

PR #61, opened 2016-11-22
don't require vsc-install, it wont work without it anyway

JensTimmerman:
requiring vsc-install will now try to install vsc-install from github, which is not always what you want? see https://github.com/hpcugent/vsc-base/issues/243


JensTimmerman:
this will break rpm creation.

PR #62, opened 2016-12-13
include Álvaro in authors list

boegel:


lib/vsc/install/shared_setup.py
```diff
@@ -121,6 +122,7 @@ def _log(self, level, msg, args):
 
 # available authors
 ag = ('Andy Georges', 'andy.georges@ugent.be')
+asg = ('Álvaro Simón García', 'alvaro.simongarcia@UGent.be')
```
alvarosimon: I'm not sure if this will work with latin-1... :smile: maybe we should remove the accents

lib/vsc/install/shared_setup.py
```diff
@@ -1,3 +1,4 @@
+# -*- coding: latin-1 -*-
```
alvarosimon: maybe `# -*- coding: utf-8 -*- `

lib/vsc/install/shared_setup.py
boegel: seems to work, tests pass... @JensTimmerman?

PR #66, opened 2017-01-13
don't cleanup setup.cfg right before creating an rpm

JensTimmerman:
for some reason when we manually generate a setup.cfg vsc_install will remove it before a call to bdist_rpm, this doesn't sound all that smart, let's not do that.

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -1158,11 +1158,6 @@ class vsc_setup(object):
                 except OSError:
                     log.error("cleanup failed for %s" % d)
 
-        for fn in ('setup.cfg',):
-            ffn = prefix + fn
-            if os.path.isfile(ffn):
-                os.remove(ffn)
-
     @staticmethod
     def sanitize(name):
         """
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -1,6 +1,6 @@
 # -*- coding: latin-1 -*-
 #
-# Copyright 2011-2016 Ghent University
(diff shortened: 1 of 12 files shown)


lib/vsc/install/shared_setup.py
```diff
@@ -1158,11 +1158,6 @@ def cleanup(self, prefix=''):
                 except OSError:
                     log.error("cleanup failed for %s" % d)
 
-        for fn in ('setup.cfg',):
-            ffn = prefix + fn
-            if os.path.isfile(ffn):
-                os.remove(ffn)
```
boegel: @JensTimmerman since it's here so explicitly, there must be a reason for it?

Have you tried figuring out why it was added?

Also, what made you make this change? Why is this a problem all of a sudden?

JensTimmerman: It was added in the initial commit...

Turns out when you create your own setup.cfg and tell vsc_install to not create one (using createsetupcfg=False) it will not create one, but first delete the one you created in this cleanup, before going to the create_rpm step, which is not what you want.

If you let it create one it will create one and overwrite the one that was there anyway, so no need to remove it first.

stdweird: are you sure it will create one if one already exists without deleting it first?

JensTimmerman: https://github.com/hpcugent/vsc-install/blob/master/lib/vsc/install/shared_setup.py#L1411

stdweird: @JensTimmerman thx for looking it up. indeed, always creates its own file.

PR #67, opened 2017-01-31
added python-noevn.sh

JensTimmerman:


boegel:
@JensTimmerman Not sure I like the `python-noenv` name, since this exposes it easily to users (e.g. on tab completion of `python`).

Maybe we should give it a more obscure name?

Also , no tests?

JensTimmerman:
no this is fine

PR #72, opened 2017-04-27
Always update urls and dependency links, even if name is set

JensTimmerman:
When name was set in PACKAGE vsc-install would not auto detect the url, and not decide if this was a private package and not updating the dependency_links.
This fixes this.


JensTimmerman:
needed for e.g. https://jenkins8.ugent.be/job/hpcugentgithubugentjenkinsfile/job/icinga-checks/job/PR-94/

Issue #63, opened 2016-12-13
auto publish master builds to pypi


JensTimmerman:
when master builds correctly and a version was bumped, (perhaps by checking if this is a specific tag?) the packages should be built and pushed to pypi.
options:
 - have jenkins do this (in the jenkinsfile?)
 - have travis do this (https://docs.travis-ci.com/user/deployment/pypi/)
 - have vsc-install do this itself (using python and passwords from config files)

stdweird:
there's already the `vsc_release` command, which now only prnits what you should run, incl tag in github,
requires config file for pypi and some env variable with the github secret

boegel:
If you want to let Travis do this, we might as well switch to Travis entirely.

Please don't start mixing Jenkins & Travis for a project.

JensTimmerman:
@boegel agreed

boegel:
EasyBuild bootstrap got broken again after merge of #66 because no new release of `vsc-install` was published on PyPI (fixed manually now).

Any ETA for getting this fixed?

Maybe we should have a separate Jenkins jobs that reacts to `vsc-install` merges to `master` which verifies if a corresponding release was pushed to PyPI, so we don't forget about it until this issue is handled?

JensTimmerman:
I'll start working on this (because I'm often bitten by this)

I'll create a new setup.py command: jenkins_run
This should combine the test and vsc_release command, in that it will run the tests, if successful check `if os.environ['GIT_BRANCH'] == 'master'` and if so run the commands that are now generated by vsc_release.

We could then create 1 Jenkinsfile that only does: set up setuptools, install vsc-install, call `python setup.py jenkins_run`
If in the future anything else needs to happen in tests or on jenkins, no need to change the jenkinsfile everywhere, we can fix this in vsc-install.

any remarks?

stdweird:
a procedure and/or a new `vsc_something` call to revoke a release?

JensTimmerman:
Is it really possible to revoke a release? You mean reupload the previous version with a higher version number?

boegel:
You can remove the sources from PyPI, but it should be done with great care, since you can never re-upload again for the same version.

Removing sources is frowned upon, all software has bugs, just do a new release with a bumped version...

JensTimmerman:
see also https://github.com/hpcugent/vsc-install/issues/7
This command should also run clusterbuildrpm and publish releases on github

PR #79, opened 2018-02-27
fix for catastrophic backtracking

JensTimmerman:


lib/vsc/install/shared_setup.py
```diff
@@ -1115,7 +1115,12 @@ def run(self):
             fullname = self.distribution.get_fullname()
 
             url = self.distribution.get_url()
-            gh_reg = re.search(r'^.*?://([^/]*github[^/]*)/', url)
+            # this is a regex with catastrophic backtracking, so limit the length of url. This takes 10 secs to process on
```
stdweird: line too long

boegel:
I tried pushing this update to PyPI (since not having it there breaks stuff, see #63), but it failed:

```
error: Upload failed (400): 'stijn.deweirdt@ugent.be;andy.georges@ugent.be;jens.timmerman@ugent.be' is an invalid value for Author-email. Error: Invalid email address see https://packaging.python.org/specifications/core-metadata
```

Seems like we should be using `<mail1>, <mail2>` as format, see https://packaging.python.org/specifications/core-metadata/#author-email-optional.

I'll issue a PR to try and fix.

PR #80, opened 2018-02-28
fix format for Author-email field

boegel:
cfr. https://packaging.python.org/specifications/core-metadata/#author-email-optional

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -148,7 +148,7 @@ URL_GHUGENT_HPCUGENT = 'https://github.ugent.be/hpcugent/%(name)s'
 
 RELOAD_VSC_MODS = False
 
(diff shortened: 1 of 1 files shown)


boegel:
publishing new `vsc-install` releases to PyPI is currently broken with:

```
error: Upload failed (400): 'stijn.deweirdt@ugent.be;andy.georges@ugent.be;jens.timmerman@ugent.be' is an invalid value for Author-email. Error: Invalid email address see https://packaging.python.org/specifications/core-metadata
```

This should be fixed ASAP because PyPI releases are strictly required to avoid things breaking, cfr. #63.

boegel:
OK, that indeed fixed it, vsc-install 0.10.29 is now on PyPI: https://pypi.python.org/pypi/vsc-install/0.10.29

PR #82, opened 2018-03-09
 fix prospector in correct place

JensTimmerman:


lib/vsc/install/shared_setup.py
```diff
@@ -1252,15 +1252,6 @@ def parse_target(self, target, urltemplate=None):
...
         new_target.update(vsc_setup_klass.SHARED_TARGET)
 
-        if sys.version_info < (2, 7):
-            # py26 support dropped in 0.8, and the old versions don't detect enough
-            log.info('no prospector support in py26 (or older)')
-        else:
-            log.info('adding prospector to tests_require')
-            tests_requires = new_target.setdefault('tests_require', [])
-            # add fixed pylint-django here as workaround for https://github.com/landscapeio/prospector/issues/245
-            tests_requires.extend(['prospector >= 0.12.1', 'pylint-django == 0.9.1'])
```
boegel: @JensTimmerman Can you clarify why you're dropping this?

What will happen now on systems running Python 2.6?

JensTimmerman: this code is never used? See below where it is replicated and is used.

boegel: OK

PR #86, opened 2018-05-23
Pylintfix:  depedencies break in new pylint and pyflakes

JensTimmerman:
fix pylint and pyflakes to older version

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -1359,7 +1359,7 @@ class vsc_setup(object):
         else:
             log.info('adding prospector to tests_require')
             tests_requires = new_target.setdefault('tests_require', [])
-            tests_requires.extend(['prospector >= 0.12.1', 'pylint-django == 0.9.1'])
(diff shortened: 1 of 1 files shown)


lib/vsc/install/shared_setup.py
```diff
@@ -1381,7 +1381,7 @@ def parse_target(self, target, urltemplate=None):
         else:
             log.info('adding prospector to tests_require')
             tests_requires = new_target.setdefault('tests_require', [])
-            tests_requires.extend(['prospector >= 0.12.1', 'pylint-django == 0.9.1'])
+            tests_requires.extend(['prospector >= 0.12.1', 'pylint<2.0.0', 'pyflakes<1.7.0,>=1.5.0'])
```
boegel: @JensTimmerman OK to drop the requirement on specific version of `pylint-django`?

Also: missing version bump?

JensTimmerman: yes, this workaround is no longer needed since it was fixed, see https://github.com/PyCQA/prospector/issues/245

itkovian:
Version bump?

PR #88, opened 2018-06-07
check for + fix correct shebang in scripts in check_header

boegel:


JensTimmerman:
the check wasn't really needed because vsc-install will always write it out when installing.

boegel:
Well, it's good to catch these issues before installing, no?

Also, explain me how https://github.ugent.be/hpcugent/vsc-filesystems-quota/pull/40 happened then?

lib/vsc/install/headers.py
```diff
@@ -180,12 +180,39 @@ def check_header(filename, script=False, write=False):
...
+                     file_ext, filename)
+
+        changed = shebang != orig_shebang
+
+    if orig_shebang is not None:
         # original position
-        header_end_pos += 1 + len(shebang)  # 1 is from splitted newline
+        header_end_pos += 1 + len(orig_shebang)  # 1 is from splitted newline
 
         if 'python' in shebang and shebang != SHEBANG_ENV_PYTHON:
```
boegel: @JensTimmerman It only changed the shebang if i) if there was a shebang (see also `shebang is not None` above), ii) if it has `python` in it and it wasn't the expected `#!/usr/bin/env python`

JensTimmerman: that's only the code that does the checking and auto fixing of headers.

I'm saying that vsc-install adds the shebang sat install time, not at test time

https://github.com/hpcugent/vsc-install/blob/be3a2b0261cd497b4dd9e286242f94becd46cc83/lib/vsc/install/shared_setup.py#L621

JensTimmerman: but currently only for python scripts, not for bash scripts. might need to be extended for that.

JensTimmerman: I seem to be confused, that line also only does the replacement of env to non env
there used to be code around somewhere that added shebangs, but I can't seem to find it

boegel: If it was there, https://github.ugent.be/hpcugent/vsc-filesystems-quota/pull/40 wouldn't have happened.

Also, there was no test that checked whether Python scripts with missing shebang was being corrected, now there is.

itkovian: The python scripts had no shebang line after installation (and did not have any in the repo) when https://github.ugent.be/hpcugent/vsc-filesystems-quota/pull/39 was merged.

itkovian:
+1.

PR #95, opened 2019-02-13
tweak egg_info & sdist directly in self.cmdclass in vsc_bdist_rpm.run (HPC-7552)

### LINKED ISSUE #94: "python setup.py bdist_rpm" doesn't replace shebang lines in scripts with recent setuptools
When `vsc-install` is used on top of a recent version of `setuptools` (e.g. `38.4.0`), shebang lines in scripts are no longer replaced as expected when building an RPM with `python setup.py bdist_rpm`.

For some reason, the `vsc_sdist_rpm.make_release_tree` does not get triggered, even though `vsc_bdist_rpm.run` does get used based on the the output of `python setup.py bdist_rpm`, and `vsc.sdist.make_release_tree` gets triggered as well.

Maybe the injection of `vsc_sdist_rpm` into `cmdclass['sdist']` (which is done in `vsc_bdist_rpm.run`) happens too late because of changes in recent `setuptools` versions?

boegel:
(fixes #94)

With old `setuptools` versions, `klass.SHARED_TARGET['cmdclass']` happens to be a *reference* to `self.cmdclass`, so changing that also changes `self.cmdclass`.

This is no longer the case in recent `setuptools` versions however, because the arguments passed to the `setup` function are being copied into a new dictionary: see https://github.com/pypa/setuptools/blob/v38.4.0/setuptools/__init__.py#L113.

Hence, `klass.SHARED_TARGET['cmdclass']` is no longer a reference to `self.cmdclass`, and any changes made to it have no effect whatsoever since `SHARED_TARGET` is only used in `parse_target`, which is used to construct the arguments for the call to the `setup` function.

This basically worked by accident I think, doesn't look like we were intentionally avoiding to use `self.cmdclass`...

~~Note: this is untested, but I expect it to work after deep diving into the `setuptools` & `distutils` code and comparing differences between `setuptools` versions 0.9.8 and 38.4.0...~~
Verified by tweaking `vsc-install` version on `regice` and (locally) rebuilding `vsc-jobs` RPM (without pushing it to `geodude` repo).

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -155,7 +155,7 @@ URL_GHUGENT_HPCUGENT = 'https://github.ugent.be/hpcugent/%(name)s'
 
 RELOAD_VSC_MODS = False
 
-VERSION = '0.11.5'
+VERSION = '0.11.6'
 
 log.info('This is (based on) vsc.install.shared_setup %s' % VERSION)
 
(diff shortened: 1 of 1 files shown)


boegel:
For some reason unclear to me, downgrading `setuptools` back to 0.9.8 on `regice` is insufficient to avoid this bug, probably because some other Python package that is pulled in from the `openstack` repo causes trouble somehow? 🤷‍♂️ 

With this fix in place, the problem *is* resolved (on top of setuptools 0.9.8 at least, to be confirmed on `setuptools` 38.4.0)...

PR #100, opened 2019-03-25
Add python3 recommendations to README.md HPC-7605

hajgato:


README.md
```diff
@@ -218,6 +218,162 @@ except (ExceptionOne, ExceptionTwo) ...
...
+os.chmod(foo, 0700)
+```
+=>
+```python
+os.chmod(foo, 0o700)
+```
+
+Import star module level
+------------------------
+Do not import \*, be more specific. If it is impossible, import it in the top level.
```
stdweird: this will still trigger some other warning. can you add the ' #noqa: ..' or whatever test to ignore here as well?

README.md
```diff
@@ -218,6 +218,162 @@ except (ExceptionOne, ExceptionTwo) ...
...
+```python
+print foo,bar
+```
+=>
+
+```python
+import from __future__ import print_function
+
+
+print (foo,bar)
```
stdweird: no spaces between function name and `(` (actually, only mention the oneliner below)

README.md
```diff
@@ -218,6 +240,127 @@ except (ExceptionOne, ExceptionTwo) ...
 
 (espcially when used like `except A, B:` which should be `except (A, B):`.
 
+Fixing print statement
+----------------------
+
+Use the oneliner:
+```bash
+find lib bin -name '*.py' | xargs futurize -w -f libfuturize.fixes.fix_print_with_import -n
```
wdpypere: maybe add this depends on python-future

PR #101, opened 2019-04-10
Add support for VUB

wpoely86:
This makes it possible for us to also use `vsc-install`. 

It writes a different header based on the url of the project.


lib/vsc/install/headers.py
```diff
@@ -117,6 +117,21 @@ def get_header(filename, script=False):
     return header, shebang
 
 
+def gen_license_header_brussel(license_template):
+    """
+    This adjusts the license template for use in Brussel
+    :param license_template: a license template
+    """
+    # replace university name
```
stdweird: no, template it for all sites (using templating ;)

wpoely86: done, but it will now fail if the url doesn't contain either `hpcugent` or `sisc-hpc`. Should I by default fallback to `hpcugent`?

stdweird: hmmmm, lets try without the fallback

wpoely86:
@stdweird can you also build the rpm and push it to geodude? Thanks.

wdpypere:
@wpoely86 done

PR #106, opened 2019-05-13
add "inconsistent-return-statements" to the whitelist

wdpypere:
I was bitten by this in a script with nasty side effects. Can we enable this?
to prevent following example like functions returning None:

```
def test:
    if something:
        return integer
```

### DIFF
--- a/lib/vsc/install/commontest.py
+++ b/lib/vsc/install/commontest.py
@@ -116,7 +116,7 @@ PROSPECTOR_WHITELIST = [
     'redefined-builtin',
     'print-statement',  # use print() and from future import __print__ instead of print
     'metaclass-assignment',  # __metaclass__ doesn't exist anymore in python3
-    'inconsistent-return-statements', # Either all or none return statements in a function should return an expression.
+    'inconsistent-return-statements', # Either all or no return statements in a function should return an expression.
 ]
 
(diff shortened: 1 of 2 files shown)


stdweird:
euhm, i'd rather not. this will affect a lot of code, and forces to do all error reporting via raise etc etc

lib/vsc/install/commontest.py
```diff
@@ -116,6 +116,7 @@
     'redefined-builtin',
     'print-statement',  # use print() and from future import __print__ instead of print
     'metaclass-assignment',  # __metaclass__ doesn't exist anymore in python3
+    'inconsistent-return-statements', # Either all or none return statements in a function should return an expression.
```
itkovian: `all or no`

itkovian:
I wonder how many issues this will raise :p

wdpypere:
I understand why we would not add this. I just got bitten by it and wanted to propose it.

wdpypere:
for the record. I do think it is bad practice and should be avoided. :wink:

itkovian:
I am all in favour, mind.

wdpypere:
Actually, exception handling is not needed at all. A simple `return None` would do. It just forces you to be explicit in your entire function.

PR #107, opened 2019-05-22
specify long_description_content_type based on extension of README file

boegel:
general fix that replaces https://github.com/hpcugent/vsc-mympirun/pull/154

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -156,7 +156,7 @@ URL_GHUGENT_HPCUGENT = 'https://github.ugent.be/hpcugent/%(name)s'
 
 RELOAD_VSC_MODS = False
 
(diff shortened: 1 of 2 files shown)


boegel:
@stdweird With this included, I was able to publish the new `vsc-mympirun` version to PyPI using:

```
python setup.py sdist
twine upload --verbose dist/vsc-mympirun-4.1.8.tar.gz
```

boegel:
`twine` is now the recommend tool for uploading Python packages to PyPI, so I've updated the output of `vsc_release`; for example for current `vsc-mypirun` version 4.1.8:

```
$ python setup.py vsc_release
...
INFO: Register with pypi
# Run commands below to upload to PyPI (testpypi False)
python setup.py register sdist
twine upload --verbose dist/vsc-mympirun-4.1.8.tar.gz

PR #108, opened 2019-05-22
Add bh as author

hajgato:


lib/vsc/install/shared_setup.py
```diff
@@ -156,7 +157,7 @@ def _log(self, level, msg, args):
 
 RELOAD_VSC_MODS = False
 
-VERSION = '0.12.5'
+VERSION = '0.12.6'
```
boegel: Hey! That's my version number for #107 ;)

hajgato: not anymore

PR #109, opened 2019-08-08
Make sure that grep never uses colors

wpoely86:
We had a user that had:
```bash
export GREP_OPTIONS='--color=always'
```
in his `.bash_profile` and it breaks python-stripped-env.

### DIFF
--- a/bin/python-stripped-env
+++ b/bin/python-stripped-env
@@ -33,6 +33,9 @@ set -e -u
 # actual python binary that we'll call
 PYTHON=/usr/bin/python
 
+# avoid weird issue with grep (for example, colored output)
+unset GREP_OPTIONS
+
 # only actually print debug info if $PYTHON_STRIPPED_ENV_DEBUG is defined
 function debug {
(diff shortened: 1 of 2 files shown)


bin/python-stripped-env
```diff
@@ -43,14 +43,14 @@ function debug {
 debug "which python: $(which python)"
 
 # unset all $LD_* environment variables
-for env_var in $(env | egrep '^LD_' | cut -f1 -d'='); do
+for env_var in $(env | egrep --color=never '^LD_' | cut -f1 -d'='); do
```
stdweird: just unset GREP_OPTIONS, much safer and future proof

wpoely86: Nope, from the man page:

```
       GREP_OPTIONS
              This variable specifies default options to be placed in front of any explicit options.  As this causes problems when writing portable scripts, this feature will be removed
              in a future release of grep, and grep warns if it is used.  Please use an alias or script instead.
```

stdweird: so unsetting it is actually recommended...

wpoely86: ok

PR #111, opened 2019-09-26
Ignore user site directory in python-stripped-env

wpoely86:
Things in `~/.local/lib/python2.7/` still get picked up by `python-stripped-env`

It seems like python 3 has something even better: `-I`.

### DIFF
--- a/bin/python-noenv.sh
+++ b/bin/python-noenv.sh
@@ -25,5 +25,9 @@
 # along with vsc-install. If not, see <http://www.gnu.org/licenses/>.
 #
 ### END OF HEADER
-# the -s makes sure the user site package is also ignored
+#
+# env -i: start with an empty environment
+# -s: ignore $HOME/.local/python*/site-packages
+# see https://docs.python.org/2/using/cmdline.html#cmdoption-s
+# and https://docs.python.org/2/library/site.html#site.USER_SITE
 env -i /usr/bin/python -s "$@"
--- a/bin/python-noenv.sh
+++ b/bin/python-noenv.sh
@@ -25,4 +25,5 @@
 # along with vsc-install. If not, see <http://www.gnu.org/licenses/>.
 #
 ### END OF HEADER
-env -i /usr/bin/python "$@"
+# the -s makes sure the user site package is also ignored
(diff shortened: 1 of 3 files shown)


bin/python-stripped-env
```diff
@@ -57,5 +57,5 @@ for env_var in $(env | grep '^PYTHON' | cut -f1 -d'=' | grep -v PYTHON_STRIPPED_
     debug "> \$$env_var was: ${!env_var}"
 done
 
-debug "$PYTHON -E $(echo "$@")"
-$PYTHON -E "$@"
+debug "$PYTHON -s -E $(echo "$@")"
```
boegel: let's add some comments here to explain what these options do:

```python
# -s: ignore $HOME/.local/python*/site-packages
# see https://docs.python.org/2/using/cmdline.html#cmdoption-s
# and https://docs.python.org/2/library/site.html#site.USER_SITE
# -E: ignore $PYTHONPATH and $PYTHONHOME
# see https://docs.python.org/2/using/cmdline.html#cmdoption-e

File: bin/python-noenv.sh [hunk 2]
```diff
@@ -25,4 +25,5 @@
 # along with vsc-install. If not, see <http://www.gnu.org/licenses/>.
 #
 ### END OF HEADER
-env -i /usr/bin/python "$@"
+# the -s makes sure the user site package is also ignored
```
boegel: Add same comment as in in python-stripped-env, and add a comment to explain why `env -i` is used?

wpoely86:
@stdweird any objections on this change?

stdweird:
@wpoely86 no, seems fine `PYTHONUSERBASE` variable might also be intersting (not sure what problems you are facing)

wpoely86:
The original problem was the logging in to your clusters was very slow for me because I had some stuff under `~/.local/lib/python2.7` which slowed down `vsc_env` a lot (it took 20 seconds to login).

I clean up but we should avoid it in general because the users might still do this too.

stdweird:
@wpoely86 no, this PR makes perfect sense, should have been like this from the start.

PR #112, opened 2019-11-19
improved python naming control via VSC_RPM_PYTHON variable

stdweird:


boegel:
@stdweird needs more love, the existing test for `sanitize` fails, and you should enhance it to also cover the case where `$VSC_RPM_PYTHON` is set to `2` or `3`

PR #114, opened 2019-11-22
add support for generating Jenkinsfile & tox.ini using "python -m vsc.install.ci --force" (HPC-7603)

boegel:
* updated `Jenkinsfile` & added `tox.ini` included here were auto-generated with `python -m vsc.install.ci --force`
* a `revision` tag is included in both auto-generated `Jenkinsfile` and `tox.ini` files, so the tests can verify they're up-to-date (or need to be updated via `python -m vsc.install.ci --force`); I'm deliberately *not* using the `vsc-install` version in the revision tag (not all future `vsc-install` updates will result in changes to the generated `Jenkinsfile` or `tox.ini`)
* Travis CI config (`.travis/yml`) was updated (manually) to use `tox` & test with both Python 2.7 and 3.6 (which helped a lot in making sure this all worked before opening the PR and letting Jenkins run the tests)
* tests with Python 3.6 will fail for `vsc-install` in Travis due to compatibility issues that haven't been fixed yet, but `tox` is configured to allow failing tests for now
* tests with Python 3.6 will fail in Jenkins for all `vsc-*` packages for now (since the required `python3.6` is missing on `jenkins8`), but `tox` has been configured to ignore that (to avoid having to move all `vsc-*` packages to the new (not-ready-yet) Jenkins server once this gets merged); this will be fixed when moving `vsc-install` (& co) to the new Jenkins server that has both Python 2.7 and 3.6 available (not easy to do on current `jenkins8`)
* tests were added to `vsc.install.commontest` to verify that `Jenkinsfile` and `tox.ini` are present and up-to-date for all `vsc-*` packages (based on the revision tag)

Note: merging this will effectively require that all other `vsc-*` Python packages first run `python -m vsc.install.ci --force` to re-generate the `Jenkinsfile` (& `tox.ini`), and that they're added to the repository, otherwise the tests added to `commontest` will fail.

Tests will fail everywhere as soon as the new `vsc-install` is published on PyPI...


boegel:
See also re-generated `Jenkinsfile` and `tox.ini` for `vsc-base` in https://github.com/hpcugent/vsc-base/pull/285

boegel:
`tox` installation on Jenkins fails with:

```
Moving filelock-3.0.12-py2.7.egg to /var/lib/jenkins/.local/lib/python2.7/site-packages
error: [Errno 2] No such file or directory: '/var/lib/jenkins/.local/lib/python2.7/site-packages/filelock-3.0.12-py2.7.egg'
```

I'll figure that one out later...

Jenkinsfile
```diff
@@ -1,6 +1,13 @@
...
 node {
-  stage 'checkout git'
-  checkout scm
-  stage 'test'
-  sh "python setup.py test"
-}
+    stage 'checkout git'
+    checkout scm
+    stage 'test'
+    sh 'python2.7 -V'
```
stdweird: what is this for?

Jenkinsfile
```diff
@@ -1,6 +1,13 @@
...
+    sh 'python -m easy_install -U --user tox'
+    sh 'tox -v'
+}
```
stdweird: add newline

lib/vsc/install/ci.py
```diff
@@ -0,0 +1,185 @@
...
+"""
+from __future__ import print_function
+from optparse import OptionParser
+import logging
+import os
+import sys
+
+
+JENKINSFILE = 'Jenkinsfile'
+JENKINSFILE_REVISION = 'Jenkinsfile@20191122-02'
```
stdweird: still don't understand what this is supposed to fix, but at leat the part before the `@` should be form previous constant

lib/vsc/install/ci.py
```diff
@@ -0,0 +1,185 @@
...
+TOX_INI = 'tox.ini'
+TOX_INI_REVISION = 'tox.ini@20191122-01'
```
stdweird: same remark

lib/vsc/install/ci.py
```diff
@@ -0,0 +1,185 @@
...
+
+logging.basicConfig(format="%(message)s", level=logging.INFO)
+LOG = logging.getLogger()
```
stdweird: why is this needed? `logging....` should work after previous line

lib/vsc/install/ci.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def parse_options(args):
+    """Parse options."""
+    parser = OptionParser()
+
+    parser.add_option('-f', '--force', dest='force', action='store_true', help="Use force to overwrite existing files")
```
stdweird: why is this optional? you are always supposed to use it....

lib/vsc/install/ci.py
```diff
@@ -0,0 +1,185 @@
...
+    lines.extend([
+        "commands = python setup.py test",
+        '',
+        # allow failing tests in Python 3, for now...
+        '[testenv:%s]' % py3_env,
+        "ignore_outcome = true"
+    ])
+
+    txt = '\n'.join(lines)
+    write_file(tox_ini, txt, force=force)
```
stdweird: do the write in the main and return text here, it will be sooooo much easier to unittest

lib/vsc/install/ci.py
```diff
@@ -0,0 +1,185 @@
...
+    cwd = os.getcwd()
+    jenkinsfile = os.path.join(cwd, JENKINSFILE)
+
+    def indent(line, level=1):
+        """Indent string value with level*4 spaces."""
+        return ' ' * 4 * level + line
+
+    test_cmds = [
+        # make very sure Python 2.7 is available,
+        # since tox ignores failures due to missing Python versions
```
stdweird: well, since we configured tox to to ignore this (add a ref to the config entry above)

lib/vsc/install/ci.py
```diff
@@ -0,0 +1,185 @@
...
+
+    cwd = os.getcwd()
+    jenkinsfile = os.path.join(cwd, JENKINSFILE)
+
+    def indent(line, level=1):
+        """Indent string value with level*4 spaces."""
+        return ' ' * 4 * level + line
+
+    test_cmds = [
+        # make very sure Python 2.7 is available,
```
stdweird: so when we start tetsing py36, we either add another version test or we remove the tox ignore missing?

lib/vsc/install/ci.py
```diff
@@ -0,0 +1,185 @@
...
+
+    lines.extend([
+        "commands = python setup.py test",
+        '',
+        # allow failing tests in Python 3, for now...
+        '[testenv:%s]' % py3_env,
+        "ignore_outcome = true"
+    ])
+
+    txt = '\n'.join(lines)
```
stdweird: add anotehr newline, to always end with newline

lib/vsc/install/ci.py
```diff
@@ -0,0 +1,185 @@
...
+        '',
+        "node {",
+        indent("stage 'checkout git'"),
+        indent("checkout scm"),
+        indent("stage 'test'")
+    ]
+    lines.extend([indent("sh '%s'" % c) for c in test_cmds])
+    lines.append('}')
+
+    txt = '\n'.join(lines)
```
stdweird: add another newline

lib/vsc/install/ci.py
```diff
@@ -0,0 +1,185 @@
...
+    write_file(jenkinsfile, txt, force=force)
```
stdweird: do this in main

lib/vsc/install/commontest.py
```diff
@@ -273,3 +275,23 @@ def test_prospector(self):
...
         self.assertFalse(failures, "prospector failures: %s" % pprint.pformat(failures))
+
+    def test_jenkinsfile(self):
+        """Test whether Jenkinsfile is in place, and was auto-generated with correct revision."""
+
+        self.assertTrue(os.path.exists('Jenkinsfile'))
+        with open('Jenkinsfile') as fh:
+            txt = fh.read()
+
+        regex = re.compile(r'^// \[revision: %s\]$' % re.escape(JENKINSFILE_REVISION), re.M)
```
stdweird: this is pointless imho. just generate the expected text and compare the generated text with text from file (it gives you a nice diff too).

lib/vsc/install/commontest.py
```diff
@@ -273,3 +275,23 @@ def test_prospector(self):
...
+        self.assertTrue(regex.search(txt), "Pattern '%s' found in tox.ini: %s" % (regex.pattern, txt))
+
+    def test_tox_ini(self):
+        """Test whether tox.ini is in place, and was auto-generated with correct revision."""
+
+        self.assertTrue(os.path.exists('tox.ini'))
+        with open('tox.ini') as fh:
+            txt = fh.read()
+
+        regex = re.compile(r'^# \[revision: %s\]$' % re.escape(TOX_INI_REVISION), re.M)
```
stdweird: same remark

test/ci.py
```diff
@@ -0,0 +1,187 @@
...
+    from cStringIO import StringIO  # Python 2
+except ImportError:
+    from io import StringIO  # Python 3
+
+import vsc.install.ci
+from vsc.install.ci import JENKINSFILE_REVISION, TOX_INI_REVISION, gen_jenkinsfile, gen_tox_ini, write_file
+from vsc.install.testing import TestCase
+
+
+def read_file(path):
```
stdweird: why is this a separate function?

test/ci.py
```diff
@@ -0,0 +1,187 @@
...
+class CITest(TestCase):
+    """License related tests"""
+
+    def setUp(self):
+        """Test setup"""
+        super(CITest, self).setUp()
+
+        self.cwd = os.getcwd()
+        self.tmpdir = tempfile.mkdtemp()
+        self.tmpdir_name = os.path.basename(self.tmpdir)
```
stdweird: what is this for?

test/ci.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+class CITest(TestCase):
+    """License related tests"""
+
+    def setUp(self):
+        """Test setup"""
+        super(CITest, self).setUp()
+
+        self.cwd = os.getcwd()
```
stdweird: i'm fine with moving this to `testing.py`, it's used a lot elsewhere too

stdweird: incl next lines

test/ci.py
```diff
@@ -0,0 +1,187 @@
...
+        self.tmpdir = tempfile.mkdtemp()
+        self.tmpdir_name = os.path.basename(self.tmpdir)
+
+    def tearDown(self):
+        """Test cleanup"""
+        shutil.rmtree(self.tmpdir)
```
stdweird: and also this and next line

test/ci.py
```diff
@@ -0,0 +1,187 @@
...
+        os.chdir(self.cwd)
+
+        super(CITest, self).tearDown()
+
+    def run_function(self, function, *args, **kwargs):
```
stdweird: wtf is this? generate the expected text and compare it with a static file in test/data
if you want to test `write_file`, just use mock?

test/ci.py
```diff
@@ -0,0 +1,187 @@
...
+
+        function(*args, **kwargs)
+
+        stdout = stringio.getvalue()
+        vsc.install.ci.LOG.handlers = orig_handlers
+
+        return stdout
+
+    def test_gen_jenkinsfile(self):
+        """Test generating of Jenkinsfile."""
```
stdweird: please make this simpler.

Jenkinsfile
```diff
@@ -1,6 +1,13 @@
...
+// This file was automatically generated using 'python -m vsc.install.ci -f'
+// DO NOT EDIT MANUALLY
+
 node {
-  stage 'checkout git'
-  checkout scm
-  stage 'test'
-  sh "python setup.py test"
-}
+    stage 'checkout git'
```
boegel: Note to self: use new syntax here (see https://github.com/hpcugent/quattor/blob/master/Jenkinsfile)

(thx @wdpypere)

Jenkinsfile
boegel: To be very sure that we're still testing with Python 2.7, since we're configuring `tox` to ignore missing interpreters (so we can get this merged on `jenkins8` where Python 3.6 is not available yet.

lib/vsc/install/ci.py
boegel: Because I really don't like blindly overwriting files by default...
Usually the `Jenkinsfile` and `tox.ini` will be in a git repo, but not always (although I admit it's unlikely).
If you think this is pointless, then I can drop this option and just blindly overwrite.

In practice you'll always run `python -m vsc.install.ci -f`, indeed.

lib/vsc/install/ci.py
boegel: Once we start running the tests on a Jenkins setup that has both Python 2.7 and 3.6, we can drop the `skip_missing_interpreters = True`.

test/ci.py
boegel: Just to avoid copy-pasting the same lines in multiple places.

test/ci.py
boegel: This is mostly about capturing the logging to stdout (the stuff the user gets to see, at least)

test/ci.py
boegel: Please clarify what you mean. ;)

I guess moving the `write_file` to `main` will help quite a bit already...

lib/vsc/install/ci.py
stdweird: i think this is pointless because it is meant for git repos, so any accidental overwrites are still not that bad

test/ci.py
stdweird: `read_file(filename)` vs `open(filename).read()`? anyway, try to avoid testing with files. you can typically mock everything you need

test/ci.py
stdweird: euh, not sure what he added value of unittesting the log output is, but whatever i guess

test/ci.py
stdweird: it's a lot of code to compare 1 static piece of text with a generated one.

lib/vsc/install/ci.py
boegel: I removed it

lib/vsc/install/ci.py
boegel: dropped the `--force`, blindly overwriting now...

lib/vsc/install/ci.py
boegel: done

lib/vsc/install/ci.py
boegel: fixed

lib/vsc/install/ci.py
boegel: fixed

lib/vsc/install/ci.py
boegel: done

lib/vsc/install/commontest.py
boegel: removed

lib/vsc/install/commontest.py
boegel: removed

test/ci.py
boegel: removed, no longer needed

test/ci.py
boegel: removed, no longer needed, but this could indeed be added in `vsc.install.testing` if it pops up again

test/ci.py
boegel: removed

test/ci.py
boegel: should be a lot better now...

Jenkinsfile
boegel: fixed in cdd0740

lib/vsc/install/ci.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def write_file(path, txt):
+    """Write specified contents to specified path."""
+    try:
+        with open(path, 'w') as handle:
+            handle.write(txt)
+        logging.info("Wrote %s", path)
+    except (IOError, OSError) as err:
+        raise IOError("Failed to write %s in %s: %s" % (path, os.getcwd(), err))
```
stdweird: why is the `in %s` referrring to cwd? if path is relative, prerfix it with cwd before writing

lib/vsc/install/ci.py
```diff
@@ -0,0 +1,163 @@
...
+        # python2.7 must always be available though, see Jenkinsfile
+        "skip_missing_interpreters = true",
+        '',
+        '[testenv]',
+    ]
+
+    if install_vsc_install:
+        lines.extend([
+            # use easy_install rather than pip to install vsc-install dependency
+            # (vsc-* packages may not work when installed with pip due to use of namespace package vsc.*)
```
stdweird: what is this about? why would one need pip or anything like that?

lib/vsc/install/commontest.py
```diff
@@ -273,3 +275,32 @@ def test_prospector(self):
 
         failures = run_prospector(self.setup.REPO_BASE_DIR)
         self.assertFalse(failures, "prospector failures: %s" % pprint.pformat(failures))
+
+    def test_jenkinsfile(self):
+        """Test whether Jenkinsfile is in place, and was auto-generated with correct revision."""
```
stdweird: no revision anymore

lib/vsc/install/commontest.py
```diff
@@ -273,3 +275,32 @@ def test_prospector(self):
...
+            current_jenkinsfile = fh.read()
+
+        expected_jenkinsfile = gen_jenkinsfile()
+
+        error_msg = "Contents of %s does not match expected contents, " % JENKINSFILE
+        error_msg += "you should run 'pyton -m vsc.install.ci' again to re-generate %s" % JENKINSFILE
+        self.assertEqual(current_jenkinsfile, expected_jenkinsfile, error_msg)
+
+    def test_tox_ini(self):
+        """Test whether tox.ini is in place, and was auto-generated with correct revision."""
```
stdweird: no revision anymore

lib/vsc/install/ci.py
```diff
@@ -0,0 +1,163 @@
...
+            'commands_pre = python -m easy_install -U vsc-install',
```
stdweird: why not do this for vsc-install too? it will be there anyway

lib/vsc/install/ci.py
stdweird: ok, think i understand. it 's a bit tricky to bootstrap possibly, but other than that, should be ok. i sortoffrogit how this used to work before...

stdweird:
some minor remarks, looks ok overall

lib/vsc/install/ci.py
boegel: `pip` is the recommended tool to install Python packages (and I think `setuptools`/`easy_install` is even deprecated), but we can't use it because of the way we do the `vsc.*` namespace currently (which is not fully supported by `pip` for some bizarre reason)

lib/vsc/install/ci.py
boegel: This installs the latest `vsc-install` from PyPI, which is not the one we want to test.

But I guess the one in the current directory will take precedence anyway, so I guess it doesn't hurt (and it simplifies things a bit).

boegel:
@stdweird OK now?

PR #126, opened 2020-01-02
fix running vsc-install tests with Python 3 (+ update copyright lines)

boegel:


stdweird:
this is surpirsingly simple (well, lot of code is not unittested)

boegel:
@stdweird That's probably largely thanks to #83, but yeah, this probably doesn't mean yet that `vsc-install` is to be considered fully compatible with Python 3 just yet...

PR #127, opened 2020-01-03
add tests for vsc.install.testing (taken from vsc-base) + fix issues with Python 3

boegel:
Tests for `vsc.install.testing` were still located in `vsc-base` after move from `vsc.utils.testing`, but they belong here.

Some tests in `vsc-base` are failing with Python 3 because of the issues fixed here.

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -159,7 +159,7 @@ URL_GHUGENT_HPCUGENT = 'https://github.ugent.be/hpcugent/%(name)s'
 
 RELOAD_VSC_MODS = False
 
-VERSION = '0.14.0'
+VERSION = '0.14.2'
 
 log.info('This is (based on) vsc.install.shared_setup %s' % VERSION)
 log.info('(using setuptools version %s located at %s)' % (setuptools.__version__, setuptools.__file__))
--- a/lib/vsc/install/testing.py
+++ b/lib/vsc/install/testing.py
@@ -142,7 +142,7 @@ class TestCase(OrigTestCase):
         try:
             call(*args, **kwargs)
             str_kwargs = ['='.join([k, str(v)]) for (k, v) in kwargs.items()]
-            str_args = ', '.join([str(x) for x in args] + str_kwargs)
+            str_args = ', '.join(list(map(str, args)) + str_kwargs)
             self.assertTrue(False, "Expected errors with %s(%s) call should occur" % (call.__name__, str_args))
         except error as err:
             msg = self.convert_exception_to_str(err)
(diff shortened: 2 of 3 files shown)


itkovian:
LGTM.

lib/vsc/install/shared_setup.py
```diff
@@ -159,7 +159,7 @@ def _log(self, level, msg, args):
 
 RELOAD_VSC_MODS = False
 
-VERSION = '0.14.0'
+VERSION = '0.14.2'
```
stdweird: why the jump? anything missing?

lib/vsc/install/testing.py
```diff
@@ -142,7 +142,7 @@ def assertErrorRegex(self, error, regex, call, *args, **kwargs):
         try:
             call(*args, **kwargs)
             str_kwargs = ['='.join([k, str(v)]) for (k, v) in kwargs.items()]
-            str_args = ', '.join(map(str, args) + str_kwargs)
+            str_args = ', '.join([str(x) for x in args] + str_kwargs)
```
stdweird: i'd prefer the py23 module with proper funtion for this.

test/testing.py
```diff
@@ -0,0 +1,124 @@
...
+        self.assertEqual(self.count_logcache('error'), 3)
+
+        self.reset_logcache('error')
+        self.assertEqual(self.count_logcache('error'), 0)
+
+        myerror("Error")
+        myerror("Moar error")
+        self.assertEqual(self.count_logcache('error'), 2)
+
+
```
stdweird: remove all things below

lib/vsc/install/shared_setup.py
boegel: See commit msg, I accidentally published a `0.14.1` to PyPI (with stuff not included in this PR). I removed that version again from PyPI shortly after (and you can't recycle version numbers on PyPI, rightfully so).

lib/vsc/install/testing.py
boegel: As discussed in Slack, I'm not in favor of adding a module to `vsc-install` which results in making it a runtime dependency, doesn't make sense imho...

If so, we might as well merge `vsc-install` and `vsc-base` into one huge library (and maybe collapse `vsc-utils` into it while we're at it).

Generally speaking, using `map` isn't recommend anyway I think, especially not where having an iterator doesn't make much sense (like here).

test/testing.py
boegel: fixed in 49f98bd

lib/vsc/install/testing.py
stdweird: again, the runtime dep is only temporary, until we drop py2 (ie probably in 2021). merging vsc-utils is a no-go due to its dependencies

PR #132, opened 2020-01-09
add support for opting in to also running shellcheck in generated Jenkinsfile

### LINKED ISSUE #131: support shellcheck in Jenkinsfile
we have places where we test bash in our Jenkinsfile doing things like:
```
    stage('test bash') {
        sh './shellcheck bin/*.sh'
    }
```
with the new generated Jenkinsfile this no longer works. Is there a way to add this?

boegel:
fixes #131 (see also https://github.ugent.be/hpcugent/quattor-host-mngmt-tools/pull/11)

### DIFF
--- a/lib/vsc/install/ci.py
+++ b/lib/vsc/install/ci.py
@@ -51,6 +51,7 @@ VSC_CI = 'vsc-ci'
 VSC_CI_INI = VSC_CI + '.ini'
 
 JIRA_ISSUE_ID_IN_PR_TITLE = 'jira_issue_id_in_pr_title'
+RUN_SHELLCHECK = 'run_shellcheck'
 
 logging.basicConfig(format="%(message)s", level=logging.INFO)
 
@@ -122,6 +123,7 @@ def parse_vsc_ci_cfg():
     """Parse vsc-ci.ini configuration file (if any)."""
     vsc_ci_cfg = {
         JIRA_ISSUE_ID_IN_PR_TITLE: False,
+        RUN_SHELLCHECK: False,
     }
 
     if os.path.exists(VSC_CI_INI):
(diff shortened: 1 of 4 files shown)


lib/vsc/install/ci.py
```diff
@@ -177,11 +183,24 @@ def indent(line, level=1):
...
+
+    if vsc_ci_cfg[RUN_SHELLCHECK]:
+        # see https://github.com/koalaman/shellcheck#installing
+        shellcheck_url = 'https://storage.googleapis.com/shellcheck/shellcheck-latest.linux.x86_64.tar.xz'
+        lines.extend([indent("stage ('shellcheck') {"),
+            indent("sh 'curl --silent %s --output - | tar -xJv'" % shellcheck_url, level=2),
+            indent("sh 'cp shellcheck-latest/shellcheck .'", level=2),
+            indent("sh 'rm -r shellcheck-latest'", level=2),
+            indent("sh './shellcheck --version'", level=2),
+            indent("sh './shellcheck bin/*.sh'", level=2),
```
stdweird: we can make the default (to use shellcheck or not) by checking the presence of `.sh` files in bin?

boegel: I prefer controlling the contents via `vsc-ci.cfg`, there may be cases where we don't actually want `shellcheck` to run on `bin/*.sh` (e.g. an "urgent" fix in a repository that has shell scripts, but hasn't been using `shellcheck` yet).

Also, having a single place that controls what goes in `Jenkinsfile`/`tox.ini` is way better than having various magic ways of enabling/disabling stuff imho...

stdweird: well, you can always turn it off. but i don;t care
we can alwasy change it later

PR #141, opened 2020-02-10
Only inject dependency version when a equal is allowed

wpoely86:
Inject dependency links with expliciet version is only useful when `==`,
`>=` or `<=` is used. If there is not a `=`, the dependency links is
pointless.


wpoely86:
It's more complex then I thought. A dependency_link needs a proper version attached to it if you also ask for a version of the dependency, else it's ignored:
https://setuptools.readthedocs.io/en/latest/setuptools.html#id14

boegel:
@wpoely86 Add that URL to the comment?

PR #140, opened 2020-02-07
stick to mock < 4.0 when using Python 2

boegel:

PR #142, opened 2020-02-11
Fix dependency_links injection for real

wpoely86:
This should be the final fix for #139 and #140 

One major issue: a split on `>` matches both `>` and `>=`...

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -1433,15 +1433,15 @@ class vsc_setup(object):
             new_target['tests_require']):
 
             # see https://docs.python.org/3/reference/expressions.html#comparisons
-            split_re = r'(\s)?([<>]=?|[=!]=)(\s)?'
-            check_whitespace = re.search(split_re, dependency)
+            split_re = re.compile(r'(\s)?([<>]=?|[=!]=)(\s)?')
+            check_whitespace = split_re.search(dependency)
 
             if check_whitespace and (check_whitespace.group(1) is None or check_whitespace.group(3) is None):
                 raise ValueError("Missing spaces around comparison operator in '%s'" % dependency)
 
             if dependency.startswith('vsc'):
-                dep_name = re.split(split_re, dependency)[0]
-                dep_name_version = re.sub(split_re, '-', dependency)
+                dep_name = split_re.split(dependency)[0]
+                dep_name_version = split_re.sub('-', dependency)
                 # if you specify any kind of version on a dependency, the dependency_links also needs a version or
                 # else it's ignored: https://setuptools.readthedocs.io/en/latest/setuptools.html#id14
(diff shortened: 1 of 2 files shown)


test/shared_setup.py
```diff
@@ -219,3 +219,37 @@ def test_parse_target(self):
...
+                'vsc-config >= 2.0.0',
+                'vsc-accountpage-clients',
+                'vsc-base > 1.0.0'
+            ],
+        }
+        setup = vsc_setup()
+        # this is needed to pass the tests on Travis: travis will clone vsc-install through
+        # https and it will be marked as non private repo, causing the dependency_links to be injected
+        # with git+https, which is not correct in this test case.
+        setup.private_repo = True
```
boegel: @wpoely86 Then make this conditional, only set this when the `$TRAVIS` environment variable is defined?

wpoely86: it's going to be True in most cases (like on my laptop or Jenkins)?

lib/vsc/install/shared_setup.py
```diff
@@ -1431,28 +1428,25 @@ def parse_target(self, target, urltemplate=None):
             ]
         else:
             urls = [('github.com', 'git+https://')]
+
         for dependency in set(new_target['install_requires'] + new_target['setup_requires'] +
             new_target['tests_require']):
+
+            # see https://docs.python.org/3/reference/expressions.html#comparisons
+            split_re = r'(\s)?([<>]=?|[=!]=)(\s)?'
```
boegel: @wpoely86 Since you're using this same regex in multiple places, why not compile it here and then use it below?

```python
split_re = re.compile(r'(\s)?([<>]=?|[=!]=)(\s)?')

File: lib/vsc/install/shared_setup.py [hunk 3]
```diff
@@ -1431,28 +1428,25 @@ def parse_target(self, target, urltemplate=None):
...
+            check_whitespace = re.search(split_re, dependency)
```
boegel:
```python
check_whitespace = split_re.search(dependency)

File: lib/vsc/install/shared_setup.py [hunk 4]
```diff
@@ -1431,28 +1428,25 @@ def parse_target(self, target, urltemplate=None):
...
+
+            if check_whitespace and (check_whitespace.group(1) is None or check_whitespace.group(3) is None):
+                raise ValueError("Missing spaces around comparison operator in '%s'" % dependency)
+
             if dependency.startswith('vsc'):
-                dep = dependency.split(' ')[0]
-                depversion = ''
-                # if you specify any kind of version on a dependency, the depedency_links also needs a version or
+                dep_name = re.split(split_re, dependency)[0]
```
boegel:
```python
dep_name = split_re.split(dependency)[0]

PR #144, opened 2020-03-26
stop using deprecated assertEquals, use assertEqual instead

boegel:
Doesn't require a version bump, since it's only in the tests, and doesn't change anything functionally...

### DIFF
--- a/test/shared_setup.py
+++ b/test/shared_setup.py
@@ -213,9 +213,9 @@ class TestSetup(TestCase):
         klass.SHARED_TARGET['cmdclass']['easy_install'] = object
         new_target = setup.parse_target(package)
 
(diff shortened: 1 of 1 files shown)


itkovian:
I'm not sure I agree.

boegel:
@itkovian With what? The lack of a version bump?

The code covered by the version (which doesn't include `test/`) literally hasn't changed.

In addition, `assertEquals` and `assertEqual` are *exactly* the same thing, except for the name, so it's literally a perfectly equivalent renaming operation in `test/`, nothing changes except that we avoid that the tests start failing whenever `assertEquals` has disappeared...

itkovian:
Allee, tis hoe.

PR #146, opened 2020-03-27
upgrade pip before using 'pip install' to install tox in Jenkinsfile

boegel:


lib/vsc/install/ci.py
```diff
@@ -201,10 +201,17 @@ def indent(line, level=1):
         'python2.7 -V',
     ]
 
+
     if vsc_ci_cfg[PIP_INSTALL_TOX]:
-        test_cmds.append('pip install --ignore-installed --user tox')
+        test_cmds.extend([
+            'pip install --user --upgrade pip',
```
wpoely86: why not for pip3?

boegel: Because I haven't run into problems yet with `pip3` on `jenkins10`... :)

PR #149, opened 2020-04-30
also add --install-scripts option to 'pip install' used in Jenkinsfile

boegel:
@wdpypere This is the fix you need for https://github.ugent.be/hpcugent/icinga-checks/pull/153

### DIFF
--- a/lib/vsc/install/ci.py
+++ b/lib/vsc/install/ci.py
@@ -201,19 +201,30 @@ def gen_jenkinsfile():
         'python2.7 -V',
     ]
 
+    if vsc_ci_cfg[INSTALL_SCRIPTS_PREFIX_OVERRIDE]:
+        pip_args = '--install-option="--install-scripts={envdir}/bin" '
+        easy_install_args = '--script-dir={envdir}/bin '
+    else:
+        pip_args = ''
+        easy_install_args = ''
 
     if vsc_ci_cfg[PIP_INSTALL_TOX]:
+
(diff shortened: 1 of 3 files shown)


lib/vsc/install/ci.py
```diff
@@ -201,19 +201,30 @@ def indent(line, level=1):
         'python2.7 -V',
     ]
 
+    if vsc_ci_cfg[INSTALL_SCRIPTS_PREFIX_OVERRIDE]:
+        pip_args = '--install-option="--install-scripts={envdir}/bin" '
+        easy_install_args = '--script-dir={envdir}/bin '
+    else:
```
itkovian, line 207: I'd prefer this be the default. It's hard to see that this will exist in the below if.

itkovian:
~~~~
pip_args = ''
easy_install_args = ''
if vsc_ci_cfg[INSTALL_SCRIPTS_PREFIX_OVERRIDE]:
    pip_args += '--install-option="--install-scripts={envdir}/bin" '
    easy_install_args += '--script-dir={envdir}/bin '
~~~~

lib/vsc/install/ci.py
```diff
@@ -201,19 +199,26 @@ def indent(line, level=1):
         'python2.7 -V',
     ]
 
+    pip_args, easy_install_args = '', ''
```
itkovian, line 202: Ugly, but I'll allow it :p

PR #151, opened 2020-05-05
use .eggs.py2 or .eggs.py3, depending on Python version being used

boegel:
If this approach is better, I can also roll back the changes done in #150 in this PR, since they're not needed anymore...

It will still require to update `.gitignore` everywhere though, to add `.eggs.py2` and `.eggs.py3`.

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -76,7 +76,7 @@ except ImportError:
 # Test that these are matched by a .gitignore pattern
 GITIGNORE_PATTERNS = ['.pyc', '.pyo', '~']
 # .gitnore needs to contain these exactly
-GITIGNORE_EXACT_PATTERNS = ['.eggs', '.eggs.py2', '.eggs.py3']
+GITIGNORE_EXACT_PATTERNS = ['.eggs*']
 
 # private class variables to communicate
 # between VscScanningLoader and VscTestCommand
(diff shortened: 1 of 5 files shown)


lib/vsc/install/shared_setup.py
```diff
@@ -260,6 +261,29 @@ def _fvs(msg=None):
...
+        if not os.path.exists(egg_cache_dir_pyver):
+            os.rename(egg_cache_dir, egg_cache_dir_pyver)
+    except OSError as err:
+        raise OSError("Failed to rename %s to %s: %s" % (egg_cache_dir, egg_cache_dir_pyver, err))
+
+    return egg_cache_dir_pyver
+
+
+setuptools.dist.Distribution.get_egg_cache_dir = get_egg_cache_dir_pyver
+setuptools.dist.Distribution.get_egg_cache_dir = get_egg_cache_dir_pyver
```
wdpypere, line 284: @boegel why do this twice?

lib/vsc/install/shared_setup.py
```diff
@@ -75,7 +76,7 @@
 # Test that these are matched by a .gitignore pattern
 GITIGNORE_PATTERNS = ['.pyc', '.pyo', '~']
 # .gitnore needs to contain these exactly
-GITIGNORE_EXACT_PATTERNS = ['.eggs']
+GITIGNORE_EXACT_PATTERNS = ['.eggs', '.eggs.py2', '.eggs.py3']
```
stdweird, line 79: `.eggs*`?

lib/vsc/install/shared_setup.py
boegel: fat fingers, will fix ;)

lib/vsc/install/shared_setup.py
boegel: Yeah, good point, might as well...

PR #155, opened 2020-05-25
limit more test dependencies (zipp, configparser)

wpoely86:
Some more packages are python 3 only.

Also do not limit the version in case of python 3.

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -160,7 +160,7 @@ URL_GHUGENT_HPCUGENT = 'https://github.ugent.be/hpcugent/%(name)s'
 
 RELOAD_VSC_MODS = False
 
-VERSION = '0.15.10'
+VERSION = '0.15.11'
 
 log.info('This is (based on) vsc.install.shared_setup %s' % VERSION)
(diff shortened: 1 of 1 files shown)


lib/vsc/install/shared_setup.py
```diff
@@ -1438,19 +1438,24 @@ def parse_target(self, target, urltemplate=None):
...
-            tests_requires = new_target.setdefault('tests_require', [])
-            new_target['tests_require'] = [x for x in tests_requires if 'prospector' not in x]
-        else:
+            tests_requires = [x for x in tests_requires if 'prospector' not in x]
+        elif sys.version_info < (3, 0):
             log.info('adding prospector to tests_require')
-            tests_requires = new_target.setdefault('tests_require', [])
+            # limit version to avoid pulling in python 3.x only versions
+            tests_requires.append('zipp < 2.0')
+            tests_requires.append('configparser < 5.0')
```
boegel, line 1450: Are these (indirect) test dependencies? Via which pkg are these required?

wpoely86: As far as I can tell, pydocstyle is the issue. It pulls in configparser without any restriction.

zipp is unclear. I think it's through pyflakes -> importlib_metadata -> zipp.

PR #156, opened 2020-05-26
add support for running 'pip install' from $HOME in Jenkinsfile

boegel:


wdpypere:
nice, thanks!

lib/vsc/install/ci.py
```diff
@@ -204,21 +206,30 @@ def indent(line, level=1):
         pip_args = '--install-option="--install-scripts={envdir}/bin" '
         easy_install_args = '--script-dir={envdir}/bin '
 
+    # run 'pip install' commands in $HOME (rather than in repo checkout) if desired
+    if vsc_ci_cfg[HOME_INSTALL]:
+        install_cmd = "cd $HOME && pip install"
```
wpoely86, line 211: why not use `pip2`? On my Fedora, `pip` means means `pip3`.

boegel: We're actually using `pip3` pretty much everywhere now, and I prefer not changing stuff unless we have a good reason for it...

wpoely86: Why do you have a `elif vsc_ci_cfg[PIP3_INSTALL_TOX]:` ?

It would make it clear what you are doing. Now it's up to the system where this is run to know what `pip` means.

boegel: Using `pip3` is opt-in (for now), see https://github.com/hpcugent/vsc-install/#use-pip3-to-install-tox

PR #161, opened 2020-07-06
ignore build directory when running prospector

FKD13:


boegel:
@itkovian Any idea why this was needed?

stdweird:
@boegel i wasted a few hours of debugging on why prospector kept failing, and it was checking code in an older build dir (soe leftover of some setup.py command @FKD13 ran)

boegel:
@stdweird No version bump?

PR #163, opened 2020-07-16
small bugfix

wpoely86:
Fixes error introduced by #160


lib/vsc/install/shared_setup.py
```diff
@@ -659,7 +659,7 @@ def make_release_tree(self, base_dir, files):
                 for fn in scripts:
                     # includes newline
                     first_line = _read(os.path.join(base_dir, fn), read_lines=True)
-                    if pyshebang_reg.search(first_line):
+                    if pyshebang_reg.search(first_line[0]):
```
stdweird, line 662: make the `first_line` the actual first line? (ie. move the `[0]` a few lines up)

PR #164, opened 2020-07-27
make .git extension optional for repository URL

lexming:
Repository `https` URLs do not necessarily end in `.git`. For instance, `git clone https://github.com/hpcugent/vsc-install` works and the remote will be registered without `.git`.


boegel:
lgtm, thanks @lexming!

PR #165, opened 2020-08-10
don't install tox with --user in Jenkinsfile, use --prefix instead

boegel:
This is done to avoid that the tests run for different repositories step on each others toes on `jenkins10`, by not using a shared tox installation in `$HOME/.local`...

### DIFF
--- a/lib/vsc/install/ci.py
+++ b/lib/vsc/install/ci.py
@@ -215,10 +215,13 @@ def gen_jenkinsfile():
     if vsc_ci_cfg[PIP_INSTALL_TOX]:
         pip_args += '--ignore-installed --user'
 
+        # tox requires zipp: require zipp < 3.0 since newer version are Python 3 only
+        tox = '"zipp<3.0" tox'
+
         test_cmds.extend([
             install_cmd + ' --user --upgrade pip',
             # make sure correct 'pip' installation is used
-            'export PATH=$HOME/.local/bin:$PATH && %s %s tox' % (install_cmd, pip_args),
+            'export PATH=$HOME/.local/bin:$PATH && %s %s %s' % (install_cmd, pip_args, tox),
         ])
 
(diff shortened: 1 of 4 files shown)


lib/vsc/install/ci.py
```diff
@@ -212,27 +212,36 @@ def indent(line, level=1):
...
         easy_install_args += '-U --user'
         test_cmds.append('%s %s tox' % (install_cmd, easy_install_args))
 
-    # make sure 'tox' command installed with --user is available via $PATH/$PYTHONPATH
-    test_cmds.append('export PATH=$HOME/.local/bin:$PATH && tox -v -c %s' % TOX_INI)
+    test_cmds.extend([
+        # make sure 'tox' command installed is available by updating $PATH
+        'export PATH=%s:$PATH && tox -v -c %s' % (os.path.join(prefix, 'bin'), TOX_INI),
+        # clean up tox installation
+        'rm -r $PWD/.vsc-tox',
```
wpoely86, line 243: shouldn't use let `git clean` deal with that?

boegel: Maybe, but we know what we need to remove, so might as well keep it specific?

wpoely86: You're not sure that nothings else lingers around? Why not be thorough and let `git clean` do everything?

wpoely86:
Missing version bump

lib/vsc/install/ci.py
boegel: We run `git clean -fxd` on Jenkins before running the tests, see `Jenkinsfile`.

This cleanup is probably not even needed (since if the tests are run again, `git clean` should clean it up), but I feel it's cleaner to clean up something temporary rather than leaving it linger.

wdpypere:
LGTM

PR #171, opened 2020-09-02
--install-scripts install option should only be used in tox.ini, not in Jenkinsfile

boegel:
The `--install-scripts` option is only needed to override the one specified in `setup.cfg`, but the `pip install` commands in `Jenkinsfile` in repos where this is an issue are run from `$HOME` to ignore the `setup.cfg`

Using `--install-option` when installing `tox` in `Jenkinsfile` causes problems, because it trigers `pip` to not download pre-built wheels from PyPI:
```
UserWarning: Disabling all use of wheels due to the use of --build-options / --global-options / --install-options
```

The side effect of this is that `tox` dependencies like `virtualenv` need to be installed from source, which results in problems like:
```
setuptools >= 41 required to build
```

This is basically a bug in `vsc-install`, there's no point in passing the `--install-scripts` option when installing `tox` in `Jenkinsfile`...

### DIFF
--- a/test/ci.py
+++ b/test/ci.py
@@ -285,4 +285,4 @@ class CITest(TestCase):
         pip3_regex = re.compile('pip3 install')
         pip3_install_scripts = pip_install_scripts.replace('pip ', 'pip3 ')
         expected_jenkinsfile = pip3_regex.sub(pip3_install_scripts, EXPECTED_JENKINSFILE_PIP3_INSTALL_TOX)
-        self.assertEqual(gen_jenkinsfile(), expected_jenkinsfile)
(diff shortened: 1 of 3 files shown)


test/ci.py
```diff
@@ -285,4 +285,4 @@ def test_install_scripts_prefix_override(self):
         pip3_regex = re.compile('pip3 install')
         pip3_install_scripts = pip_install_scripts.replace('pip ', 'pip3 ')
         expected_jenkinsfile = pip3_regex.sub(pip3_install_scripts, EXPECTED_JENKINSFILE_PIP3_INSTALL_TOX)
```
wdpypere, line 287: this line is no longer needed?

boegel: indeed, cleaned up in https://github.com/hpcugent/vsc-install/pull/172

PR #176, opened 2020-12-09
Accept all content types for README files already in code

lexming:
There is already code in `vsc-install` checking for those README file formats recommended by pypi. However, currently the only allowed README file is `README.md`. This PR takes the existing list in `readme_content_types` and uses it to check for existing README files as well (by order of preference).

The list of allowed README files becomes:
* `README.md`, `README.rst`, `README.txt`, `README`


lib/vsc/install/shared_setup.py
```diff
@@ -1565,12 +1581,15 @@ def prepare_rpm(self, target):
...
+        # Locate the readme and store its path
+        self.readme = self.locate_readme()
+
         # Add (default) and excluded_pkgs_rpm packages to SHARED_TARGET
         # the default ones are only the ones with a __init__.py file
         # therefor we regenerate self.package files with the excluded pkgs as extra param
         self.package_files = self.files_in_packages(excluded_pkgs=pkgs)
         _fvs('prepare_rpm').SHARED_TARGET['packages'] = self.generate_packages()
-        self.build_setup_cfg_for_bdist_rpm(target)
+        self.build_setup_cfg_for_bdist_rpm(target, self.readme)
```
stdweird, line 1592: why do you pass a an attribute to a method?

lib/vsc/install/shared_setup.py
```diff
@@ -1316,6 +1325,23 @@ def get_license(self, license_name=None):
         log.info("Found license name %s and classifier %s", lic_short, data[1])
         return lic_short, data[1]
 
+    def locate_readme(self):
+        """
+        Returns path to README file in the package from allowed options in README global
+        The highest option in README global is preferred
```
stdweird, line 1331: fyi, the order is README random. but not sure it matters much

stdweird:
jenkins failures
```
======================================================================

ERROR: test_parse_target (test.shared_setup.TestSetup)

Test for parse target

  File "/var/lib/jenkins/jobs/hpcugent_github.com/jobs/vsc-install/branches/PR-176/workspace/test/shared_setup.py", line 217, in test_parse_target

    new_target = setup.parse_target(package)

  File "lib/vsc/install/shared_setup.py", line 1403, in parse_target

    readmetxt = _read(self.readme)

AttributeError: 'vsc_setup' object has no attribute 'readme'


======================================================================

ERROR: test_parse_target_dependencies (test.shared_setup.TestSetup)

Test injecting dependency_links in parse_target

  File "/var/lib/jenkins/jobs/hpcugent_github.com/jobs/vsc-install/branches/PR-176/workspace/test/shared_setup.py", line 243, in test_parse_target_dependencies

    new_target = setup.parse_target(package)

  File "lib/vsc/install/shared_setup.py", line 1403, in parse_target

    readmetxt = _read(self.readme)

AttributeError: 'vsc_setup' object has no attribute 'readme'


Ran 46 tests in 22.083s


FAILED (errors=2)

```

lib/vsc/install/shared_setup.py
lexming: The problem is that `build_setup_cfg_for_bdist_rpm` is a static method, so it does not have access to object attributes. I did not want to change the scope of the method in case something breaks somewhere else, so I just passed the attribute to the function. I agree that it's not pretty though.
https://github.com/hpcugent/vsc-install/pull/176/commits/d5629a46df9c3e812547e56c5dd649bf3a8e1c8e changes the approach and uses the `long_description` in target to store the readme file path.

lib/vsc/install/shared_setup.py
lexming: you are right, I was not thinking about Python < 3.6. I moved it to a tuple so that the order of preference is always predictable.

lib/vsc/install/shared_setup.py
```diff
@@ -1608,6 +1631,13 @@ def action_target(self, target, setupfn=None, extra_sdist=None, urltemplate=None
         if do_cleanup:
             self.cleanup()
 
+        # add location of README to target long description
+        readme_file = self.locate_readme()
+        if 'long_description' in target:
+            log.info(('Ignoring the provided long_descripton.'
+                      'Please set it in the %s or disable vsc_description') % os.path.basename(readme_file))
```
stdweird, line 1638: this has been a while since i read the code, but how is can use `vsc_description` to prevent the long_description being overwritten here?

lexming: I indeed missed one check here on `vsc_description`. To avoid polluting the code with more conditionals, I have now used a slightly different approach by using a temporary attribute in `target` instead of `long_description`. This avoids any overwrites and the need of additional checks.

lib/vsc/install/shared_setup.py
```diff
@@ -1556,7 +1581,8 @@ def build_setup_cfg_for_bdist_rpm(target):
             txt.extend(["build_requires = %s" % (klass.sanitize(target['setup_requires']))])
 
         # add metadata
-        txt += ['', '[metadata]', '', 'description-file = %s' % README, '']
+        description_file = target.get('description_file', 'Unknown')
```
stdweird, line 1584: `Unknown`? why is this needed?

lib/vsc/install/shared_setup.py
```diff
@@ -1376,17 +1403,23 @@ def parse_target(self, target, urltemplate=None):
             if 'github' in urltemplate:
                 new_target['download_url'] = "%s/tarball/master" % new_target['url']
 
-        # Readme are required
-        readme = os.path.join(self.REPO_BASE_DIR, README)
-        if not os.path.exists(readme):
-            raise Exception('README is missing (was looking for %s)' % readme)
+        # README is required, use description file from target and remove it
```
stdweird, line 1406: the `is required` comment is only relevant where the the `locate_readme` is called.

stdweird: it has been a while, but how could the `description_file` be missing?

lib/vsc/install/shared_setup.py
lexming: Just a precaution in case `build_setup_cfg_for_bdist_rpm` would be called without going through `action_target` and `description-file` would be unset.

stdweird: if there is none, don't add the metadat lines in setup.cfg. but i prefer hard failure, people who can't be bothered to add simple Readme, well, <insert swearing of your choice> them.

lib/vsc/install/shared_setup.py
lexming: I moved that part of the comment. On the other hand, since `decription_file` is set in a different method, the fallbacks just cover the unlikely case that `parse_target` is called without going through `action_target`.

lib/vsc/install/shared_setup.py
lexming: ok, added https://github.com/hpcugent/vsc-install/pull/176/commits/dbea7c7f9447d8a71b0bf77b2eb2b89b88a348a5 to only set the metadata if `description_file` is set. On the other hand, the hard failure if README is missing is already there. It happens before this point, in `locate_readme` when its called from `action_target`, line 1350. So if there is no README, you don't get here.

lexming:
I'll try once more to get this merged. It is now synced with `master`.
@stdweird: can you tell me the new error of the tests?

stdweird:
@lexming weird it doesn't fail locally then.

```
======================================================================

FAIL: test_setup_cfg (test.shared_setup.TestSetup)

Test generating of setup.cfg.

  File "lib/vsc/install/testing.py", line 76, in assertEqual

    super(TestCase, self).assertEqual(a, b)

  File "/usr/lib64/python3.6/unittest/case.py", line 846, in assertEqual

    assertion_func(first, second, msg=msg)

AssertionError: '[bdist_rpm]' != '[bdist_rpm]\n\n[metadata]\n\ndescription-file = README.md'

- [bdist_rpm]

+ [bdist_rpm]


[metadata]


description-file = README.md



During handling of the above exception, another exception occurred:


  File "/var/lib/jenkins/jobs/hpcugent_github.com/jobs/vsc-install/branches/PR-176/workspace/test/shared_setup.py", line 324, in test_setup_cfg

    self.assertEqual(read_setup_cfg(), expected)

  File "lib/vsc/install/testing.py", line 98, in assertEqual

    raise AssertionError("%s:\nDIFF%s:\n%s" % (msg, limit, ''.join(diff[:self.ASSERT_MAX_DIFF])))

AssertionError: '[bdist_rpm]' != '[bdist_rpm]\n\n[metadata]\n\ndescription-file = README.md'

- [bdist_rpm]

+ [bdist_rpm]


[metadata]


description-file = README.md

:

DIFF:

```

test/shared_setup.py
```diff
@@ -313,7 +313,10 @@ def read_setup_cfg():
         os.chdir(self.tmpdir)
 
         # test with minimal target
-        vsc_setup.build_setup_cfg_for_bdist_rpm({})
+        target = {
+            'description_file': 'README.md',
```
stdweird, line 317: why do you need to set this. it's the default value right?

test/shared_setup.py
```diff
@@ -325,6 +328,7 @@ def read_setup_cfg():
 
         # realistic target
         target = {
+            'description_file': 'README.md',
```
stdweird, line 331: not sure why this is needed? we're not going to change all the setup.py to add the old default...

test/shared_setup.py
lexming: This is only needed in this unit test with an artificial target. In a real scenario the `description_file` attribute will be set by the call to `locate_readme()` in `action_target()`, which looks for the actual README file in the repo directory. The only case where a target would have no `description_file` is if there is no README, which will error out before reaching `build_setup_cfg_for_bdist_rpm()`.

PR #177, opened 2020-12-09
Add global with allowed remotes

lexming:
This PR makes it easier to update the allowed git remotes in `vsc-install`. It does not change current behaviour.


wpoely86:
@stdweird what's the actual reason for this requirement? Why demand that a git remote is present?

stdweird:
@wpoely86 i don't remember, would have to check what is derived from it. but we've never had a usecase that wouldnt fit it either.

wpoely86:
It's also trivial to work around or just add a dummy remote.

I don't mind but I never understood the point of it.

PR #178, opened 2021-02-26
manipulate dependencies for rpm building

stdweird:
we have started adding restrictions to the requirements to work around limitations of pypi (specifically when certain dependencies are only supported for more recent python versions). however, we now have run into issues where these restrictions cause probems when making rpms while the too recent version issue does not exist in such an environment

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -166,7 +166,7 @@ URL_GHUGENT_HPCUGENT = 'https://github.ugent.be/hpcugent/%(name)s'
 
 RELOAD_VSC_MODS = False
 
-VERSION = '0.17.2'
+VERSION = '0.17.3'
 
 log.info('This is (based on) vsc.install.shared_setup %s' % VERSION)
 log.info('(using setuptools version %s located at %s)' % (setuptools.__version__, setuptools.__file__))
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -1363,6 +1363,11 @@ class vsc_setup(object):
         new_target['license'] = lic_name
         classifiers.append(lic_classifier)
 
+        # a dict with key the new_target key to search and replace
+        #    value is a list of 2-element (pattern, replace) lists passed to re.sub
+        #    if returning value is empty, it is not added after the replacement
+        vsc_filter_rpm = target.pop('vsc_filter_rpm', {})
+
         # set name, url, download_url (skip name if it was specified)
         update = self.get_name_url(version=target['version'], license_name=lic_name)
         if 'name' in target:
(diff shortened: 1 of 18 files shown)


lib/vsc/install/shared_setup.py
```diff
@@ -1513,6 +1518,28 @@ def parse_target(self, target, urltemplate=None):
                     new_target['dependency_links'] += [''.join([git_scheme, url, '/hpcugent/', dep_name, '.git#egg=',
                                                                 dep_name_version])]
 
+        if 'VSC_RPM_PYTHON' in os.environ:
+            for key, snrs in vsc_filter_rpm.items():
+                def snr(txt):
+                    for pattern, replace in snrs:
+                        txt = re.sub(pattern, replace, txt)
```
itkovian, line 1525: If an earlier replace is matched by a later pattern, this may not go right :p

wpoely86:
Isn't this regex setup a bit overkill? I guess that you want to wipe the version requirement for the rpm? Isn't easier to add a dedicated option for that instead of copying a regex everywhere?

lib/vsc/install/shared_setup.py
```diff
@@ -1513,6 +1518,28 @@ def parse_target(self, target, urltemplate=None):
                     new_target['dependency_links'] += [''.join([git_scheme, url, '/hpcugent/', dep_name, '.git#egg=',
                                                                 dep_name_version])]
 
+        if 'VSC_RPM_PYTHON' in os.environ:
+            for key, snrs in vsc_filter_rpm.items():
+                def snr(txt):
```
boegel, line 1523: What is `snr` short for? Is it "search 'n replace"?
Oh my, please rename to something more obvious, like `search_replace` or whatever...

Making `snr` a function, and `snrs` a list of things to iterate over with `snr` (rather a list of `snr`s) is ... confusing.

lib/vsc/install/shared_setup.py
stdweird: i know. wasn't sure if you want to stack the subs or go to the next as soon as a sub modifid the content

stdweird:
@wpoely86 who knows. you might also change the versionsing, or do some other stuff. the code is simple enough imho to work with regex'es

wpoely86:
The code is good but if you're going to reuse the same regex (to wipe the version) over and over again, you might as well make it easy and keep the general regex for the weird cases?

lib/vsc/install/shared_setup.py
```diff
@@ -1513,6 +1518,28 @@ def parse_target(self, target, urltemplate=None):
                     new_target['dependency_links'] += [''.join([git_scheme, url, '/hpcugent/', dep_name, '.git#egg=',
                                                                 dep_name_version])]
 
+        if 'VSC_RPM_PYTHON' in os.environ:
```
wpoely86, line 1521: magic string? Isn't there a contant for this?

lib/vsc/install/shared_setup.py
```diff
@@ -1513,6 +1518,28 @@ def parse_target(self, target, urltemplate=None):
...
+            for key, snrs in vsc_filter_rpm.items():
```
boegel, line 1522: `snrs` -> `pat_repls` (patterns and replacements)

boegel (changes requested):

lib/vsc/install/shared_setup.py
boegel: I think it makes sense to keep going. The regular expressions are specified via a list, so you have control over the order in which they're processed.

boegel (approved):

PR #179, opened 2021-03-01
vsc_filter_rpm must run before prepare_rpm is run

stdweird:


lib/vsc/install/shared_setup.py
```diff
@@ -1544,7 +1544,7 @@ def search_replace(txt):
                         log.debug("new vsc_filter_rpm value for %s was empty, not adding it back", key)
 
         log.debug("New target = %s" % (new_target))
-        print(new_target)
+        print("new target", new_target)
```
boegel, line 1547: Why both the `log.debug` and `print`?

boegel (changes requested):

lib/vsc/install/shared_setup.py
stdweird: the inner setup.py used by the actual building of the rpm is not called by the `-v` flag.

boegel (approved):

PR #182, opened 2021-04-02
give errors about logging-not-lazy

wdpypere:
I was cleaning up some code and I think its time to enable `logging-not-lazy`, the other 2 `no-member` and `logging-too-few-args` are needed to show bugs I encountered when cleaning up things.


stdweird:
@wdpypere can you update the readme with an example errror and how to fix it?

wdpypere:
OK, I will

PR #183, opened 2021-04-02
add example on how to fix logging-not-lazy to readme

wdpypere:


README.md
```diff
@@ -226,6 +226,24 @@ def new_filter(b_b):
...
 
+logging-not-lazy
+----------------
+
+Don't use string interpolation when logging if not needed:
+
+```python
+name = 'world'
+program ='python'
+print('Hello %s! This is %s.' % (name, program))
```
stdweird, line 237: perhaps use an actual logging example? also, how does this work with py3 `f`-strings?

wdpypere: Good point, I'll fix the example.

when using something like:
```
logging.error("Hello, {1}. You are {0}.".format(AGE, NAME))
```
it will nag with a 
```
test.py:5:14: W1202: Use lazy % formatting in logging functions (logging-format-interpolation)
```

wdpypere: so we should probably enable that as well?

stdweird: and try f-stirngs too 

```python
var1 = "abc"
loggin.error(f"{var1}")

```
(or something ilke that)

stdweird: imho, the f-string logging is not so bad when scalars are used, but shouldn't be used with something else.

wdpypere: your example also fails on 
`test.py:10:0: W1203: Use lazy % formatting in logging functions (logging-fstring-interpolation)`

so I won't enable that one then.

PR #187, opened 2021-04-06
enable pylint assignment-from-no-return

wdpypere:
this warns on possible logical issues like:

```
def function_without_return():                                                                                                                               
    x = 1 + 1                                                                                                                                                
    print(x)                                                                                                                                                 
                                                                                                                                                             
y = function_without_return()  
```
```
E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
```

this is correct syntax, but is probably never what you want.

### DIFF
--- a/lib/vsc/install/commontest.py
+++ b/lib/vsc/install/commontest.py
@@ -90,6 +90,7 @@ PROSPECTOR_WHITELIST = [
     'F811',  # redefinition of unused name
     'W291',  # pep8: W291 / trailing whitespace
     'arguments-differ',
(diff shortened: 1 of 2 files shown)


wdpypere:
so vsc-install itself violates this assignment-from-no-return :)
@boegel can you validate if this is reasonable to enable, and how I should fix this error?

wdpypere:
as a side remark, enabling this does nothing for python2, only for python3.

boegel (approved):
lgtm

PR #189, opened 2021-04-07
add support for specifying custom location for scripts/binaries via setup.py

boegel:


lib/vsc/install/shared_setup.py
```diff
@@ -1552,11 +1552,20 @@ def search_replace(txt):
...
+
+        Creates [bdist_rpm] section with
             install_requires => requires
             provides => provides
             setup_requires => build_requires
 
+        Creates [metadata] section with
+            description-file => README file
+
+        Creates [install] section if needed,
```
stdweird, line 1565: only when building rpm?

boegel: No, only when `install-scripts` is specified in `setup.py`.

Or I don't understand what you want me to change :)

stdweird: the is called in a method called build_setup_cfg_for_bdist_rpm, i'm going to assume that only rus when building rpms?

boegel: Yes, and that's fine. We only need to control the location of the scripts when building an RPM, not when installing via `setup.py install`.

PR #193, opened 2021-08-03
add support for pre-installing dependencies before running tests

boegel:


test/ci.py
```diff
@@ -159,11 +160,19 @@ def test_parse_vsc_ci_cfg(self):
...
-        vsc_ini_txt = '\n'.join('%s=1' % key for key in default.keys())
-        vsc_ini_txt = re.sub('additional_test_commands=1', 'additional_test_commands=./more_tests.sh', vsc_ini_txt)
-        self.write_vsc_ci_ini(vsc_ini_txt)
+        vsc_ini_txt_lines = []
+        for key in default.keys():
+            # special treatment needed for non-boolean keys like additional_test_commands
+            if default[key] is None:
+                vsc_ini_txt_lines.append('%s=foo' % key)
+            else:
+                vsc_ini_txt_lines.append('%s=1' % key)
```
itkovian, line 169: So, 1 maps to True?

boegel: Yes, those are boolean configuration settings.

README.md
```diff
@@ -553,3 +553,32 @@ tox must be configured to inherit these packages in the test environment. This c
...
+
+Pre-installing dependencies before running tests
+------------------------------------------------
+
+Although ``vsc-install`` will automatically install all dependencies listed in ``setup.py`` prior to running the
+tests, there are cases where this doesn't work out as expected.
+Some Python packages only support being installed with ``pip install`` (for example because they use a namespace
+that is spread across multiple different Python packages, like ``fs`` and ``fs.sshfs``).
+
+You can specify Python packages that should be installed (with ``pip install``) before running the tests via
```
itkovian, line 565: So this is only for tests?

boegel: Yes. It's needed because `vsc-install` automagically installs the dependencies before running the tests, but the way that's done (via setuptools, not `pip`) is not compatible with some Python packages (like `fs.sshfs`).

For an actual installation, `fs.sshfs` can be installed some other way (not via `vsc-install`).

PR #198, opened 2022-01-03
fix dependencies for platformdirs and pylint-django

wdpypere:
also clean up some old dependencies. The cleanup works on my machine, looking forward to what Jenkins has to say.

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -1510,8 +1510,8 @@ class vsc_setup(object):
                 'pylint~=2.10.2',
                 'flake8~=3.8.4',
                 'prospector~=1.5.0.1',
-                # requires python 3.7
                 'pylint-django~=2.4.4',
+                # requires python 3.7
                 'platformdirs < 2.4.0',
                 'mock',
             ])
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -1505,7 +1505,7 @@ class vsc_setup(object):
                 'pycodestyle~=2.7.0',
                 'pylint~=2.12.2',
                 'flake8~=3.9.2',
-                'prospector',
+                'prospector~=1.5.3.1',
                 'pylint-plugin-utils < 0.7',
                 'pylint-django~=2.4.4',
                 # requires python 3.7
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -1497,19 +1497,15 @@ class vsc_setup(object):
             # soft pinning of (transitive) dependencies of prospector
             # ('~=' means stick to compatible release, https://www.python.org/dev/peps/pep-0440/#compatible-release);
             # updating these must be done in lockstep, see setup.cfg or pyproject.toml or whatever at:
-            # - https://github.com/PyCQA/pylint/blob/v2.10.2/setup.cfg
-            # - https://github.com/PyCQA/flake8/blob/3.8.4/setup.cfg
-            # - https://github.com/PyCQA/prospector/blob/1.5.0.1/pyproject.toml
+            # - https://github.com/PyCQA/pylint/blob/v2.12.2/setup.cfg
+            # - https://github.com/PyCQA/flake8/blob/3.9.2/setup.cfg
+            # - https://github.com/PyCQA/prospector/blob/1.5.3.1/pyproject.toml
             tests_requires.extend([
-                # stick to astroid < 2.8, as required by pylint 2.10.x
-                'astroid~=2.7.3',
-                # stick to pyflakes < 2.3.0, as required by prospector 1.5.0.x and flake8 3.8.x
-                'pyflakes~=2.2.0',
-                # stick to pycodestyle < 2.7.0, as required by flake8 3.8.x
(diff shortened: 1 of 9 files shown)


boegel:
@wdpypere Tests are still failing?

wdpypere:
@boegel yes, I need some time to figure it out I guess. any suggestion welcome.

wdpypere:
@boegel I finally managed to update the dependencies of vsc-install. (newer prospector, pylint, ...)
but now I get:
```======================================================================

FAIL: test_prospectorfail (test.prospectortest.ProspectorTest)

Test that whitelisted warnings actually fails

  File "/var/lib/jenkins/jobs/hpcugent_github.com/jobs/vsc-install/branches/PR-198/workspace/test/prospectortest.py", line 77, in test_prospectorfail

    self.assertFalse(undetected_tests, "\nprospector did not detect %s\n" % undetected_tests)

AssertionError: ['indexing-exception', 'metaclass-assignment', 'raising-string'] is not false : 

prospector did not detect ['indexing-exception', 'metaclass-assignment', 'raising-string']
```

I presume this is because of a newer prospector? Any tips?

wdpypere:
so with the upgrade of pylint to 2.12.2 some behaviours changed. already solved `raising-string`, need to look into the other 2.

lib/vsc/install/shared_setup.py
```diff
@@ -1497,19 +1497,19 @@ def parse_target(self, target, urltemplate=None):
...
-                'prospector~=1.5.0.1',
+                'pyflakes~=2.3.0',
+                'pycodestyle~=2.7.0',
+                'pylint~=2.12.2',
+                'flake8~=3.9.2',
+                'prospector',
+                'pylint-plugin-utils < 0.7',
+                'pylint-django~=2.4.4',
+                # requires python 3.7
+                'platformdirs < 2.4.0',
```
boegel, line 1512: Add a comment to explain why an older version is required?

Also, what required Python 3.7? newer versions of `platformdirs`?

lib/vsc/install/shared_setup.py
```diff
@@ -1497,19 +1497,19 @@ def parse_target(self, target, urltemplate=None):
...
-                # stick to pycodestyle < 2.7.0, as required by flake8 3.8.x
-                'pycodestyle~=2.6.0',
-                'pylint~=2.10.2',
-                'flake8~=3.8.4',
-                'prospector~=1.5.0.1',
+                'pyflakes~=2.3.0',
+                'pycodestyle~=2.7.0',
+                'pylint~=2.12.2',
+                'flake8~=3.9.2',
+                'prospector',
```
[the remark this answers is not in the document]
wdpypere: fair point, I'll update

lib/vsc/install/shared_setup.py
```diff
@@ -1497,19 +1497,19 @@ def parse_target(self, target, urltemplate=None):
...
+                'pylint-plugin-utils < 0.7',
```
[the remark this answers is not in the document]
wdpypere: because there are version conflicts? the same reason we lock all the others? I found most of the old comments superfluous as the statement at the top says they need to be updated in lockstep.

lib/vsc/install/shared_setup.py
wdpypere: euhm, we only test python 3.6, so if you install the latest of platformdirs it all just breaks.

lib/vsc/install/shared_setup.py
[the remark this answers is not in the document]
boegel: I understand why you removed the old comments, but when adding new restrictions like this it would be helpful to mention why they were added, so we know when they can be removed again?

Or were the comments that were there basically useless to you?

lib/vsc/install/shared_setup.py
boegel: So the comment should say "platformdirs >= 2.4.0 requires Python 3.7, so stick to older versions for running tests with Python 3.6"

lib/vsc/install/shared_setup.py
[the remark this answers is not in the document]
wdpypere: the more sane option (well, the one I used) is to remove all dependencies except prospector and add them one by one as required until you get the new set.

lib/vsc/install/shared_setup.py
wdpypere: same thing as what is already there, but more extensive :smile:  . I find it very superfluous to state that, because if you read the code line it is exactly what it states. But if you insist, I'll change.

lib/vsc/install/shared_setup.py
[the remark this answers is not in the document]
wdpypere: pylint-django is also new, you just did not notice :wink: . it get's pulled in but 2.5.0.1 conflicts with latest prospector, same with the plugins. astroid got remoted as it now just works.

lib/vsc/install/shared_setup.py
boegel: It wasn't clear to me what required Python 3.7, so just `# platformdirs >= 2.4.0 requires Python 3.7` is fine by me

wdpypere: fixed it

PR #206, opened 2022-11-25
add github action for tests + fix dependencies

wpoely86:


boegel:
@wpoely86 Looks like the tests are not as expected yet in GitHub Actions?
Cfr. https://github.com/wpoely86/vsc-install/actions

wpoely86:
You mean: your jenkins has some magic combo where it does work :wink: 

Anyway, I've almost got it.

wpoely86:
@boegel fix now: https://github.com/wpoely86/vsc-install/actions/runs/3559909081

.github/workflows/unittest.yml
```diff
@@ -0,0 +1,24 @@
+name: run python tests
+on: [push, pull_request]
+
+jobs:
+  python_unittests:
+      runs-on: ubuntu-latest
+      strategy:
+        matrix:
+          python: [2.7, 3.6, 3.7, 3.8, 3.9]
```
wdpypere, line 9: I would only test for 2.7, 3.6 and 3.9 as rhel8 uses 3.6 and rhel9 uses 3.9. But it also should not make that much difference.

wpoely86: true, I've trimmed it to 2.7, 3.6, 3.9 and 3.11.

boegel (approved):
lgtm

PR #207, opened 2022-11-25
fix tests requires

wpoely86:
This should hopefully help with https://github.com/hpcugent/vsc-base/pull/333


wdpypere:
@wpoely86 instead of changing the 3.7 requirements I would make a separate 3.9 if branch.

wpoely86:
Let's follow this up in #206

PR #211, opened 2023-01-30
Lock isort to py 3.6 supported version

itkovian:


lib/vsc/install/shared_setup.py
```diff
@@ -1524,6 +1524,7 @@ def parse_target(self, target, urltemplate=None):
                     'lazy-object-proxy < 1.8.0', # higher requires python 3.7
                     'jsonpickle < 3.0.0', # higher requires python 3.7
                     'importlib-metadata < 5.0.0' # no longer compatible with python 3.7
+                    'dataclasses', # required by isort from pylint
```
wpoely86, line 1527: please add that it's part of python since 3.7 (I think?)

itkovian: That should be there already, since it's in the > 3 and < 3,7 branch.

wdpypere:
@itkovian this won't work. the newest isort being pulled in (5.12.0) only works with python 3.8 and up. You need to lock isort, not pull in dataclasses.

wdpypere:
@itkovian version bump?

PR #212, opened 2023-02-01
add consider-using-with to prospector tests

wdpypere:
I do regard this as good practice, as the `close()` is often forgotten (one example even here in vsc-install).

you can read more about it here: https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/consider-using-with.html

But I don't know how many people will be annoyed by this.


wdpypere:
I ran into this when doing `CTRL+c` at an extremely unlucky moment :smile:

boegel (approved):
lgtm

PR #213, opened 2023-02-21
start removing python 2 support

wdpypere:
this is a first nudge in the direction of removing python2 support. it makes py3_only mandatory. it will try to give warnings and errors about python2, but I'm actually quite sure it just won't work.
Should not be merged before all known packages have been migrated.

### DIFF
--- a/lib/vsc/install/ci.py
+++ b/lib/vsc/install/ci.py
@@ -34,14 +34,7 @@ import logging
 import os
 import sys
 import yaml
-
-try:
-    # Python 3
-    import configparser
-except ImportError:
-    # Python 2
-    import ConfigParser as configparser
-
+import configparser
 from vsc.install.shared_setup import MAX_SETUPTOOLS_VERSION, vsc_setup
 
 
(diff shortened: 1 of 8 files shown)


lib/vsc/install/ci.py
```diff
@@ -150,12 +143,11 @@ def gen_tox_ini():
     # list of Python environments in which tests should be run
     envs = []
 
-    # don't run tests with Python 2 if 'py3_only' is set in vsc-ci.ini
+    # give error if 'py3_only' is set to false in vsc-ci.ini
     if not vsc_ci_cfg[PY3_ONLY]:
-        envs.append('py27')
+        logging.error("package is not py3_only. Not supported. Please set 'py3_only'. will ignore python2.")
```
stdweird, line 148: why should we even set it? it is the new default. `ci` module should even remove it, no?

lib/vsc/install/ci.py
```diff
@@ -222,12 +207,7 @@ def gen_tox_ini():
         lines.append("sitepackages = true")
 
     if not vsc_ci_cfg[PY3_TESTS_MUST_PASS]:
-        lines.extend([
-            '',
-            # allow failing tests in Python 3, for now...
-            '[testenv:%s]' % py3_env,
-            "ignore_outcome = true"
-        ])
+        logging.warning("py3_tests_must_pass not set to true in config. ignoring.")
```
stdweird, line 210: ? raise/fail hard.

vsc-ci.ini
```diff
@@ -2,3 +2,4 @@
 pip3_install_tox=1
```
stdweird, line 2: i think none of the py3 things should be here anymore? not sure pip3 is similar, but py3 is now default
(it's possible that clusterbuildrpm server still looks for it, best to check that too)

wdpypere:
@stdweird you are certainly right, all the `py3*` settings could be removed from the vsc-ci.ini. Which in practice would mean most projects can have the `vsc-ci.ini` just outright removed, except if they have jenkins, jira or github actions. Can I do that in a separate follow up pr?

stdweird:
followup PR is also fine

Issue #214, opened 2023-02-22
vsc-install should provide an option to generate github actions workflow file


wdpypere:
if `use_github_actions` is set, a github actions file should be created under: `.github/workflows/unittest.yml` and look something likefor example:

```
name: run python tests
on: [push, pull_request]

jobs:
  python_unittests:
      runs-on: ubuntu-20.04
      strategy:
        matrix:
          python: [3.6, 3.9, 3.11]
      steps:
      - name: Checkout code
        uses: actions/checkout@v2
      - name: Setup Python
        uses: actions/setup-python@v2
        with:
          python-version: ${{ matrix.python }}
      - name: install tox
        run: pip install tox
      - name: add mandatory git remote
        # 'hpcugent' is a mandatory remote for setup.py to work (because of get_name_url)
        run: git remote add hpcugent https://github.com/hpcugent/vsc-base.git
      - name: Run tox
        # Run tox using the version of Python in `PATH`
        run: tox -e py
```

boegel:
@wdpypere fixed with #215?

PR #215, opened 2023-02-22
add github actions generating

wdpypere:


lib/vsc/install/ci.py
```diff
@@ -62,19 +65,68 @@
...
         raise IOError("Failed to write %s: %s" % (path, err))
 
+def gen_github_action():
+    """
+    Generate tox.ini configuration file for github actions.
+    """
+    logging.info('[%s]', GITHUB_ACTIONS)
+    vsc_ci_cfg = parse_vsc_ci_cfg()
+
+    name_url = "https://github.com/hpcugent/vsc-install"
```
wdpypere, line 92: this won't work yet as I need to get this line dynamically for each repo, I tried the code of headers and it does not work. @boegel suggestions?

lib/vsc/install/ci.py
```diff
@@ -62,19 +65,68 @@
...
+            "DO NOT EDIT MANUALLY",
+        ]
+
+        txt = ['# ' + l for l in header]
+        yaml_content = {
+            'name': 'run python tests',
+            'on': ['push', 'pull_request'],
+            'jobs': {
+                'python_unittests': {
+                    'runs-on': 'ubuntu-20.04',
```
wpoely86, line 107: Maybe directly run it inside an rhel 8 container? That's our production environment anyway.

lib/vsc/install/ci.py
wdpypere: I just noticed the headers script is used with `REPO_BASE_DIR=$PWD`, maybe this will require the same trick.

wdpypere: well, found a way to fix it, but I have to admit I don't understand how it works.

lib/vsc/install/ci.py
wdpypere: according to https://github.com/actions/runner-images there is no rhel 8 container.

wdpypere:
fixes #214

lib/vsc/install/ci.py
```diff
@@ -62,19 +65,70 @@
...
 RUN_SHELLCHECK = 'run_shellcheck'
+ENABLE_GITHUB_ACTIONS = 'enable_github_actions'
 
 logging.basicConfig(format="%(message)s", level=logging.INFO)
 
 
 def write_file(path, txt):
     """Write specified contents to specified path."""
+
+    if not os.path.exists(os.path.dirname(path)):
```
stdweird, line 76: use `exist_ok = True` in makedirs instead of the test and the copy/pasting

lib/vsc/install/ci.py
```diff
@@ -62,19 +65,70 @@
...
+                            'python': [3.6, 3.9]
+                        }
+                    },
+                    'steps': [
+                        {'name': 'Checkout code', 'uses': 'actions/checkout@v3'},
+                        {'name': 'Setup Python', 'uses': 'actions/setup-python@v4',
+                         'with': {'python-version': '${{ matrix.python }}'}},
+                        {'name': 'install tox', 'run': 'pip install tox'},
+                        {'name': 'add mandatory git remote',
+                         'run': 'git remote add hpcugent %s.git' % name_url},
```
stdweird, line 121: ue f string

stdweird:
minor remakr, nice work @wdpypere

lib/vsc/install/ci.py
wdpypere: I will, as soon as this package is python 3 only :wink:

wdpypere:
@stdweird remarks addressed

boegel (approved):

lib/vsc/install/ci.py
wpoely86: Not a runner image, a container. Just add a line like: 
```
    runs-on: ubuntu-20.04
    container: rockylinux:8
    steps:
```
Or whatever your favorite rhel flavour is.

PR #216, opened 2023-02-23
add py39 to tox tests

wpoely86:
Let's be future proof and start testing python 3.9 also?


lib/vsc/install/shared_setup.py
```diff
@@ -291,7 +291,7 @@ def get_egg_cache_dir_pyver(self):
 
         # the original get_egg_cache_dir creates the .eggs directory if it doesn't exist yet,
         # but we want to have it versioned, so we rename it
-        egg_cache_dir_pyver = egg_cache_dir + '.py%s' % sys.version_info[0]
+        egg_cache_dir_pyver = egg_cache_dir + '.py%s' % '.'.join([str(x) for x in sys.version_info[0:2]])
```
wdpypere, line 294: is this strictly needed?

wdpypere:
@wpoely86 you will need https://github.com/hpcugent/vsc-install/pull/229 to make this build.

lib/vsc/install/shared_setup.py
wpoely86: Maybe not but it's a lot cleaner not to mix all python3 versions.

I guess it's not needed if you have a clean directory every time.

lib/vsc/install/ci.py
```diff
@@ -146,8 +146,8 @@ def gen_tox_ini():
     envs = []
 
     # always run tests with Python 3
-    py3_env = 'py36'
-    envs.append(py3_env)
+    py3_envs = ['py36', 'py39']
```
stdweird, line 149: how do we run tests without 39 in certain repos?

wpoely86: We can add it to `vsc-cfg.ini` to optionally disable it?

wpoely86:
I will revamp this one when #229 is merged

lib/vsc/install/ci.py
wdpypere: might be a good idea, same way we did it for the py3 move.

wdpypere: I'll look into it

stdweird: yes, should be dealt with same as in py3 move: first it's opt in, then mandatory with optout

test/ci.py
```diff
@@ -107,6 +107,9 @@
 EXPECTED_TOX_INI_PY36_IGNORE = """
 [testenv:py36]
 ignore_outcome = true
+
+[testenv:py39]
+ignore_outcome = true
```
wdpypere, line 112: this does not work right? the outcomes are not ignored

lib/vsc/install/ci.py
wdpypere: OK, I'll get something working

wdpypere:
superseded by https://github.com/hpcugent/vsc-install/pull/230

PR #217, opened 2023-02-23
put github workflows in rockylinux 8 container

wpoely86:
As all of our code runs under RHEL flavours, I would also run the github actions there.


wpoely86:
Nevermind, I thought this was possible. Not.

PR #218, opened 2023-03-07
fix some deprecation warnings and unclosed files in tests

wdpypere:
while trying to debug some other things these errors confused me. so fix them here.


wdpypere:
just some test fixes, so no version bump is needed afaik.

PR #219, opened 2023-03-07
more general cleanup

wdpypere:
- remove old hack which depends on deprecated optparse
 - remove some more python 2 related dependency handling
 - explain distutils.log usage a bit better
 - get rid of `is_string()` function
 - deprecate and ignore py3_only and py3_tests options, so they are no longer mandatory, and can safely be removed


wdpypere:
@stdweird the followup PR I promised a few weeks back

itkovian (approved):

boegel (changes requested):
@wdpypere missing version bump?

wdpypere:
@boegel version bumped

boegel (approved):

PR #223, opened 2023-04-27
Fix error in case of installing with pip and pyproject.toml

wpoely86:
This cause an error in github pipelines. The logger class needs to have
a doc string...

See https://github.com/wpoely86/vsc-accountpage/actions/runs/4821249134/jobs/8586871529


stdweird:
@wpoely86 pushed to pypi

PR #224, opened 2023-06-22
pin versions of virtualenv and tox compatible with Python v3.6

lexming:
We hit issue https://github.com/pypa/virtualenv/issues/2550 due to changes from https://github.com/pypa/virtualenv/pull/2548 . This means that all tests in Python 3.6 fail with latest versions of `tox` and `virtualenv`.

This PR caps the versions to those still compatible with Python 3.6.


boegel (approved):
lgtm

PR #227, opened 2023-09-07
Add real exception when logging the pattern not supported

stdweird:


lib/vsc/install/shared_setup.py
```diff
@@ -800,8 +801,9 @@ def loadTestsFromModule(self, module, pattern=None):  # pylint: disable=argument
                 try:
                     # pattern is new, this can fail on some old setuptools
                     testsuites = ScanningLoader.loadTestsFromModule(self, module, pattern)
-                except TypeError:
+                except TypeError as e:
                     log.warn('pattern argument not supported on this setuptools yet, ignoring')
+                    log.warn(f'original exception {e} {traceback.format_exc()}')
```
kwaegema, line 806: can't you use log.exception() to print the traceback?

stdweird: the `log` here is not `logging` related

wdpypere: that is indeed for next gen vsc-install without fancylogger.

PR #228, opened 2023-10-16
some pylint fixes

wdpypere:
mainly implement f-strings. Not finished by a long shot. As an unexpected side effect, about 5 % faster on my machine.


lib/vsc/install/commontest.py
```diff
@@ -36,10 +36,10 @@
 
 import optparse
 import os
-import pkg_resources
 import pprint
 import re
 import sys
+import pkg_resources
```
itkovian, line 42: Any reason not to keep this alphabetical?

wdpypere: pylint says the order is wrong

itkovian: why?

wdpypere: dependencies between the modules I assume.

itkovian: fair enough.

PR #229, opened 2023-10-24
add a workable test stack for python 3.9

wdpypere:
this works. I'll make an attempt at making the versions a bit newer, but it's dependency hell between prospector and pylint-django.

whoever said it tested and worked on py3.11 was surely wrong. (probably me)

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -1499,23 +1499,18 @@ class vsc_setup():
                 'isort < 5.11.0',
                 'zipp < 3.16', # no longer compatible with python 3.6
             ])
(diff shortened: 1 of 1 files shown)


wdpypere:
the github tests still fail because the currently published vsc-install on pypi tell it to install prospector 1.7.7 which has a bug fixed in 1.8.0 which is triggered here. this PR will change that to 1.10 and should fix that.

stdweird:
id you are convinced the tests will pas 3.9, can we als enforce them in the vsc ini whatever thingie?

wdpypere:
@stdweird yes, that is done here https://github.com/hpcugent/vsc-install/pull/216

PR #230, opened 2023-10-25
run python 3.9 tests by default, ignore outcome unless specified

wdpypere:
if `py39_tests_must_pass=1` is set in vsc-ci.ini tests will have to pass.

still running into the prospector bug though. (https://github.com/landscapeio/prospector/issues/527) it should have been fixed, so investigating. (maybe it was reintroduced)

### DIFF
--- a/lib/vsc/install/ci.py
+++ b/lib/vsc/install/ci.py
@@ -58,6 +58,7 @@ PIP3_INSTALL_TOX = 'pip3_install_tox'
 EASY_INSTALL_TOX = 'easy_install_tox'
 PY3_ONLY = 'py3_only'
 PY3_TESTS_MUST_PASS = 'py3_tests_must_pass'
+PY39_TESTS_MUST_PASS = 'py39_tests_must_pass'
 RUN_SHELLCHECK = 'run_shellcheck'
 ENABLE_GITHUB_ACTIONS = 'enable_github_actions'
 
@@ -145,9 +146,9 @@ def gen_tox_ini():
     # list of Python environments in which tests should be run
     envs = []
 
-    # always run tests with Python 3
-    py3_env = 'py36'
-    envs.append(py3_env)
+    # always run tests with Python 3.6 and 3.9
+    py3_envs = ['py36', 'py39']
+    envs.extend(py3_envs)
 
     pip_args, easy_install_args = '', ''
     if vsc_ci_cfg[INSTALL_SCRIPTS_PREFIX_OVERRIDE]:
@@ -162,6 +163,13 @@ def gen_tox_ini():
         # (setup.py requires vsc-install, which is not installed yet when 'python setup.py sdist' is run)
         "skipsdist = true",
     ]
+
+    if not vsc_ci_cfg[PY39_TESTS_MUST_PASS]:
+        lines.extend([
+            '',
+            '[testenv:py39]',
+            'ignore_outcome = true'
+        ])
 
     lines.extend([
         '',
@@ -220,6 +228,7 @@ def parse_vsc_ci_cfg():
         EASY_INSTALL_TOX: False,
         RUN_SHELLCHECK: False,
         ENABLE_GITHUB_ACTIONS: False,
+        PY39_TESTS_MUST_PASS: False,
     }
 
     deprecated_options = [PY3_ONLY, PY3_TESTS_MUST_PASS, PIP_INSTALL_TOX, PIP3_INSTALL_TOX]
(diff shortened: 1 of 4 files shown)


wdpypere:
jup, 1.8.0 fixed it, 1.8.1 removed it again.

wdpypere:
@stdweird @wpoely86 this fully enables python 3.9 tests with recent prospector and pylint.

wdpypere:
also sets vsc-install to `py39_tests_must_pass=1`

lib/vsc/install/ci.py
```diff
@@ -58,6 +58,7 @@
 EASY_INSTALL_TOX = 'easy_install_tox'
 PY3_ONLY = 'py3_only'
 PY3_TESTS_MUST_PASS = 'py3_tests_must_pass'
+PY39_TESTS_MUST_PASS = 'py39_tests_must_pass'
```
stdweird, line 61: it's probably best to rename the previous `PY3_TESTS_MUST_PASS` to `PY36_TESTS_MUST_PASS`, or make it an alias; also in the config files

wdpypere: it's only there for deprecations warning anymore. e.g. if it's set you get a deprecation warning to remove it

wdpypere: I don't follow, why would we add a `py36_tests_must_pass` parameter?

test/prospectortest.py
```diff
@@ -60,8 +62,10 @@ def test_prospectorfail(self):
             testfile_base = os.path.splitext(os.path.basename(testfile))[0].replace("_", "-")
             all_tests.append(testfile_base)
             for failure in failures:
-                if failure['location']['path'] == testfile and testfile_base in [failure['code'], failure['message']]:
-                    detected_tests.append(testfile_base)
+                # old prospector returns strings, new prospector returns Path
+                if failure['location']['path'] == testfile or failure['location']['path'] == Path(testfile):
```
stdweird, line 66: `if Path(failure['location']['path']) == Path(testfile):`

lib/vsc/install/commontest.py
```diff
@@ -193,18 +193,44 @@ def run_prospector(base_dir, clear_ignore_patterns=False):
     whitelist = list(map(re.compile, PROSPECTOR_WHITELIST))
 
     failures = []
+
+    # message.as_dict() has been removed after prospector 1.8.0
+    # same for location.as_dict()
+    # mimic old behaviour
+    def prospector_message_as_dict(message):
```
stdweird, line 200: better to do here
```
def prospector_message_as_dict(message):
    if hasattr(message, 'as_dict'):
        return message.as_deict()
    else:
       return {...}
```

lib/vsc/install/commontest.py
```diff
@@ -193,18 +193,44 @@ def run_prospector(base_dir, clear_ignore_patterns=False):
...
+        return {
+            "source": message.source,
+            "code": message.code,
+            "location": prospector_location_as_dict(message.location),
+            "message": message.message,
+        }
+    def prospector_location_as_dict(location):
```
stdweird, line 207: idem as above

lib/vsc/install/ci.py
stdweird: because we will start having repos that are only py39 compatible

wdpypere:
@stdweird remarks addressed

lib/vsc/install/commontest.py
```diff
@@ -193,18 +193,44 @@ def run_prospector(base_dir, clear_ignore_patterns=False):
...
+        if hasattr(message, "as_dict"):
+            return message.as_dict()
+        else:
+            return {
+                "source": message.source,
+                "code": message.code,
+                "location": prospector_location_as_dict(message.location),
+                "message": message.message,
+            }
+    def prospector_location_as_dict(location):
```
stdweird, line 210: newline in between

test/prospectortest.py
```diff
@@ -48,23 +50,24 @@ def test_prospectorfail(self):
...
 
         detected_tests = []
         all_tests = []
         for testfile in test_files:
             testfile_base = os.path.splitext(os.path.basename(testfile))[0].replace("_", "-")
             all_tests.append(testfile_base)
             for failure in failures:
-                if failure['location']['path'] == testfile and testfile_base in [failure['code'], failure['message']]:
+                # old prospector returns strings, new prospector returns Path
+                if Path(failure['location']['path']) == Path(testfile) or testfile_base in [failure['code'], failure['message']]:
```
stdweird, line 66: line too long?

wdpypere: it's in the test directory, so does not get checked

lib/vsc/install/commontest.py
wdpypere: added

test/prospectortest.py
wdpypere: fixed

wdpypere:
as a bonus, code now get's tested with pylance as well

PR #234, opened 2023-11-10
more cleanups

wdpypere:
encoding, fstring, redefine builtins, redefining from outer scope, cell-var-from-loop, ... lots of small fixes.


lib/vsc/install/ci.py
```diff
@@ -71,11 +71,11 @@ def write_file(path, txt):
 
     os.makedirs(os.path.dirname(path), exist_ok=True)
     try:
-        with open(path, 'w') as handle:
+        with open(path, 'w', encoding='utf8') as handle:
```
stdweird, line 74: use `Path(path).write_text(txt)`?

lib/vsc/install/ci.py
```diff
@@ -248,7 +248,7 @@ def parse_vsc_ci_cfg():
             cfgparser.read(VSC_CI_INI)
             cfgparser.items(VSC_CI)  # just to make sure vsc-ci section is there
         except (configparser.NoSectionError, configparser.ParsingError) as err:
-            logging.error(f"ERROR: Failed to parse {VSC_CI_INI}: {err}")
+            logging.error("ERROR: Failed to parse %s: %s", VSC_CI_INI, err)
```
stdweird, line 251: why did you revert here?

lib/vsc/install/commontest.py
```diff
@@ -250,7 +250,7 @@ def check_autogenerated_ci_config_file(testcase_instance, ci_cfg_fn, expected_co
     """Test whether specified CI configuration file is in place, and was auto-generated by vsc-install."""
 
     testcase_instance.assertTrue(os.path.exists(ci_cfg_fn))
-    with open(ci_cfg_fn) as fih:
+    with open(ci_cfg_fn, encoding='utf8') as fih:
```
stdweird, line 253: `Path.read`

stdweird: idem for most changes below

stdweird:
@wdpypere  up to you if you want to use Path or not. i'm ok with not using it also

lib/vsc/install/ci.py
wdpypere: so for logging pylint prefers lazy logging, for anything else f-strings.

wdpypere: I'm not sure what my opinion about that is.

stdweird: it's correct, but not really needed for error, as that is always logged.

lib/vsc/install/shared_setup.py
```diff
@@ -275,12 +277,10 @@ def _fvs(msg=None):
...
     """read a file, either in full or as a list (read_lines=True)"""
-    with open(source) as file_handle:
-        if read_lines:
-            txt = file_handle.readlines()
-        else:
-            txt = file_handle.read()
-    return txt
+    text = Path(source).read_text(encoding='utf8')
+    if read_lines:
+        return text.split()
```
stdweird, line 282: split on what? make it explicit it's on newlines

lib/vsc/install/shared_setup.py
```diff
@@ -1296,9 +1295,10 @@ def fix_range(txt):
     def get_md5sum(filename):
         """Use this function to compute the md5sum in the KNOWN_LICENSES hash"""
         hasher = hashlib.md5()
-        with open(filename, "rb") as fih:
-            for chunk in iter(lambda: fih.read(4096), b""):
-                hasher.update(chunk)
+        txt = Path(filename).read_bytes()
+        chunks = [txt[i:i+4096] for i in range(0, len(txt), 4096)]
```
stdweird, line 1299: move this in the for loop?

lib/vsc/install/shared_setup.py
wdpypere: you are right! but I forgot that python has splitlines(), I'll use that.

lib/vsc/install/shared_setup.py
```diff
@@ -1296,9 +1295,9 @@ def fix_range(txt):
     def get_md5sum(filename):
         """Use this function to compute the md5sum in the KNOWN_LICENSES hash"""
         hasher = hashlib.md5()
-        with open(filename, "rb") as fih:
-            for chunk in iter(lambda: fih.read(4096), b""):
-                hasher.update(chunk)
+        txt = Path(filename).read_bytes()
+        for chunk in [txt[i:i+4096] for i in range(0, len(txt), 4096)]:
```
stdweird, line 1299: why so complicated
```
for i in range(0, len(txt), 4096):
    hasher.update(txt[i:i+4096])
```

PR #237, opened 2023-11-14
start to get rid of dist_utils

wdpypere:
we also import log from distutils. Reccomendations are to use plain logging instead. But the fix needed for that in setuptools is only in a recent setuptools which we lock at v42...

So I can't fix the logging until we more to a newer setuptools


wdpypere:
to be complete, to fix the logging we need setuptools 61 or higher

stdweird:
we use setuptools to make packages, and recent setuptools shops distutils (or whatever setuptools needs). so trying to move to more recent setuptools is ok for me

stdweird:
see also https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html

PR #240, opened 2023-12-21
test release with py39 setuptool 53 requirement

stdweird:


stdweird:
fyi `SETUPTOOLS_USE_DISTUTILS=local` doesn't play nice with setuptools 41. it's a bit unexpected, but not relevant

PR #269, opened 2026-05-28
add support for setting `py39_only` in `vsc-ci.ini`

boegel:
required for:
- https://github.com/hpcugent/jobcli/pull/405


wdpypere:
why not just set `PY36_TESTS_MUST_PASS` to false?

boegel:
> `PY36_TESTS_MUST_PASS`

Because that implies still running the tests with Python 3.6, which doesn't make sense if we know it's no longer supported

wdpypere (approved):

